PR #2, opened 2019-01-09
Fixed crash and removed unneeded closing tag

cvium:
It crashes because it doesn't have a userId during initialization where this event is also triggered.

### DIFF
--- a/index.html
+++ b/index.html
@@ -17,7 +17,7 @@
         <div class="timeOfDay"><span class="timePrefix"></span> <span class="timeSuffix"></span></div>
     </div>
     <video id="video-player" crossorigin="anonymous" preload="auto"></video>
-    <img id="photo"></img>
+    <img id="photo">
     <div id="backdrop">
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -17,7 +17,7 @@ <h2>Select your media in Emby and play it here</h2>
         <div class="timeOfDay"><span class="timePrefix"></span> <span class="timeSuffix"></span></div>
     </div>
     <video id="video-player" crossorigin="anonymous" preload="auto"></video>
-    <img id="photo"></img>
+    <img id="photo">
```
joshuaboniface: Should use a self-closing tag (``).

cvium: Hmm, wonder why VSCode didn't complain about that.

JustAMan:
Hmm, @cvium didn't fix last remaining tag yet :)

joshuaboniface:
Ah crap, just saw the one fixed didn't realize there was another. Sorry @cvium !

PR #3, opened 2019-01-10
Submodule the shared ui components

cvium:
While only very few components are needed for this app (more will be included during future PRs), it's best to split it out.


JustAMan:
@cvium how would device id be generated now?

anthonylavado:
@JustAMan It looks like it's these lines: https://github.com/jellyfin/jellyfin-chromecast/blob/af18df82a716e75d0cdd0914b9ab61dfb3410683/components/maincontroller.js#L202-L203

I wonder though, because I don't have any experience with Chromecast, do duplicate names have any implications?

components/maincontroller.js
```diff
@@ -199,7 +199,8 @@
         data.options = data.options || {};
         var cleanReceiverName = cleanName(data.receiverName || '');
         window.deviceInfo.deviceName = cleanReceiverName || window.deviceInfo.deviceName;
-        window.deviceInfo.deviceId = cleanReceiverName ? CryptoJS.SHA1(cleanReceiverName).toString() : window.deviceInfo.deviceId;
+        // deviceId just needs to be unique-ish
+        window.deviceInfo.deviceId = cleanReceiverName ? btoa(cleanReceiverName) : window.deviceInfo.deviceId;
```
JustAMan: I wonder why we ever want to override it... looks like it's an internal property exposing serial number or something.

JustAMan: For this particular we eventually will, by trial and error (that is, if it breaks because of this change we'll know that this change was wrong 🤣 ).

cvium: Serial number or other IDs aren't accessible via js unfortunately

PR #5, opened 2019-01-11
Change to quicksand font and remove Robot imports

bfayers:
Makes it look like this:
image

If anything just to match our design elsewhere

### DIFF
--- a/css/jellyfin.css
+++ b/css/jellyfin.css
@@ -1,31 +1,10 @@
-@font-face {
-    font-family: 'Roboto';
-    font-style: normal;
-    font-weight: 300;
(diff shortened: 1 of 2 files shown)


JustAMan:
I'd like @nvllsvm to say smth as he was the one behind the whole rebranding choices.

joshuaboniface:
Quicksand is the font we picked for the logo, so I assume it should be good.

nodausenis:
> Makes it look like this:
> image
> 
> If anything just to match our design elsewhere

3934315ab4eb23035f08e131a11d7a6de5581607

nodausenis (approved):

PR #9, opened 2019-04-01
Get cast device volume/mute on init and send to client on connect

Froghut:
this results in the remote control panel in the web(android) client showing the chromecast device volume when connecting to a cast device. The device state (containing the volume/mute among other things) is sent via a "playbackstop" message which normally resets the player state. I also submittet a pr https://github.com/jellyfin/jellyfin-web/pull/222 to jellyfin-web that will fix this and not reset the volume part of the player state.

### DIFF
--- a/components/maincontroller.js
+++ b/components/maincontroller.js
@@ -275,6 +275,7 @@
             if (!isPlaying()) {
                 embyActions.displayUserInfo($scope, data.serverAddress, data.accessToken, data.userId);
             }
+            // when a client connects send back the initial device state (volume etc) via a playbackstop message
             embyActions.reportPlaybackProgress($scope, getReportingParams($scope), true, "playbackstop");
         }
         else if (data.command == 'SetVolume') {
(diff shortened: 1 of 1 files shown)


components/maincontroller.js
```diff
@@ -10,6 +10,8 @@
 
         resetPlaybackScope($scope);
         clearMediaElement();
+		window.VolumeInfo.Level = cast.receiver.media.Volume.level * 100;
```
JustAMan: pls fix tabs vs spaces here and in other places

components/maincontroller.js
```diff
@@ -273,6 +275,7 @@
             if (!isPlaying()) {
                 embyActions.displayUserInfo($scope, data.serverAddress, data.accessToken, data.userId);
             }
+            embyActions.reportPlaybackProgress($scope, getReportingParams($scope), true, "playbackstop");
```
dkanada: Could you add a comment here with the explanation? Might be useful in the future.

Froghut: yeah makes sense, done.

joshuaboniface:
This is live on the production Chromecast app now (https://apps.jellyfin.org/chromecast).

PR #10, opened 2019-06-17
Fix playback status in android app when reconnecting

Froghut:
When the android app was restarted while playback was paused on chromecast it would not show that something was playing on chromecast in the app.

### DIFF
--- a/components/maincontroller.js
+++ b/components/maincontroller.js
@@ -1,4 +1,4 @@
-define(['datetime', 'embyactions', 'browserdeviceprofile', '//www.gstatic.com/cast/sdk/libs/receiver/2.0.0/cast_receiver.js', '//www.gstatic.com/cast/sdk/libs/mediaplayer/1.0.0/media_player.js'], function (datetime, embyActions, deviceProfileBuilder) {
+﻿define(['datetime', 'embyactions', 'browserdeviceprofile', '//www.gstatic.com/cast/sdk/libs/receiver/2.0.0/cast_receiver.js', '//www.gstatic.com/cast/sdk/libs/mediaplayer/1.0.0/media_player.js'], function (datetime, embyActions, deviceProfileBuilder) {
 
     window.mediaManager = new cast.receiver.MediaManager(window.mediaElement);
     setInterval(updateTimeOfDay, 40000);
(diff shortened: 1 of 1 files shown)


components/maincontroller.js
```diff
@@ -1,4 +1,4 @@
-﻿define(['datetime', 'embyactions', 'browserdeviceprofile', '//www.gstatic.com/cast/sdk/libs/receiver/2.0.0/cast_receiver.js', '//www.gstatic.com/cast/sdk/libs/mediaplayer/1.0.0/media_player.js'], function (datetime, embyActions, deviceProfileBuilder) {
+define(['datetime', 'embyactions', 'browserdeviceprofile', '//www.gstatic.com/cast/sdk/libs/receiver/2.0.0/cast_receiver.js', '//www.gstatic.com/cast/sdk/libs/mediaplayer/1.0.0/media_player.js'], function (datetime, embyActions, deviceProfileBuilder) {
```
Froghut: No idea why this line shows as changed, I did not touch it and can't see the difference

components/maincontroller.js
```diff
@@ -149,6 +149,10 @@
 
         clearMediaElement();
 
+        window.playlist = [];
+        window.currentPlaylistIndex = -1;
+        embyActions.displayUserInfo($scope, $scope.serverAddress, $scope.accessToken, $scope.userId);
+
```
Froghut: when stopping playback reset playlist etc. so isPlaying() is not returning true

components/maincontroller.js
```diff
@@ -273,6 +277,9 @@
             if (!isPlaying()) {
                 embyActions.displayUserInfo($scope, data.serverAddress, data.accessToken, data.userId);
             }
+            else {
+                embyActions.reportPlaybackProgress($scope, getReportingParams($scope), true, reportEventType);
+            }
```
Froghut: when a client reconnects it sends the 'Identify' command, and if something is currently playing (or paused) report playbcack progress back so the UI in the app is updated correctly and shows the currently playing item as playing.

components/maincontroller.js
dkanada: probably just a newline change, can be fixed with an editor config

anthonylavado:
@Froghut Once we get that line change fixed up, we'll push it through the rest of the process.

Froghut:
I changed the file encoding to UTF8 BOM, that removed the changed first line.

anthonylavado:
@froghut oops, there's a conflict now, sorry :-(

Froghut:
This is actually from an earlier PR from me, I think the best thing would be a kind of merge of the two:
`else {
                embyActions.reportPlaybackProgress($scope, getReportingParams($scope), true, "playbackstop");
            }`
Is this something I can/should do?

PR #20, opened 2020-03-25
Remove web dependency (mostly)

YouKnowBlom:
Plan is to remove our dependency to web. Datetime and browserdeviceprofile should be fairly straight forward while fetchhelper (included in jellyfin-apiclient-javascript) will take a lot more work, probably better fit for another PR.

**Done:**
- Datetime
- Browserdeviceprofile

### DIFF
--- a/helpers.js
+++ b/helpers.js
@@ -993,3 +993,40 @@ function extend(target, source) {
 function parseISO8601Date(s, toLocal) {
     return new Date(s);
 }
+
+function getDisplayRunningTime(ticks) {
+    var ticksPerHour = 36000000000;
+    var ticksPerMinute = 600000000;
+    var ticksPerSecond = 10000000;
+
+    var parts = [];
+
+    var hours = ticks / ticksPerHour;
+    hours = Math.floor(hours);
+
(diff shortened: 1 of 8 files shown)


helpers.js
```diff
@@ -901,7 +901,7 @@ function getMiscInfoHtml(item, datetime) {
 
         if (item.Type == "Audio") {
 
-            miscInfo.push(datetime.getDisplayRunningTime(item.RunTimeTicks));
+            miscInfo.push(getDisplayRunningTime(item.RunTimeTicks));
```
dkanada: Does this method exist elsewhere?

YouKnowBlom: Good catch, I was a bit too fast in my cleaning frenzy. I'll throw it into helpers.js.

ferferga:
What if you go and use date-fns as well here? We plan to move all the datetime functions to it, so I think it makes sense to do it in Chromecast as well, as I suppose it's possible to do it, like in Web.

Cc @MrTimscampi

YouKnowBlom:
I'll have to look into that. Thanks for the heads up!

PR #23, opened 2020-04-10
Convert deviceprofilebuilder and related files to ES6 modules

YouKnowBlom:
Basically only changes all `define`s to `import` and exports all functions. Some variables are updated to reference the right imported functions and objects

### DIFF
--- a/src/components/deviceprofilebuilder.js
+++ b/src/components/deviceprofilebuilder.js
@@ -1,279 +1,296 @@
-define(["codecsupporthelper", "castdevices"], function (codecSupport, castDevices) {
-    "use strict";
+import {
+    deviceIds,
+    getActiveDeviceId
+} from "./castDevices";
+import {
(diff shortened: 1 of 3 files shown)


src/components/deviceprofilebuilder.js
```diff
@@ -1,279 +1,296 @@
...
-
-        return profile;
+    return TranscodingProfiles;
+}
+
+/**
+ * @returns {Array} Subtitle profiles.
+ */
+function getSubtitleProfiles() {
+    //TODO: Add TTML support
```
dkanada, line 250: Unrelated but spaces before comments are nice.

YouKnowBlom: I agree, I'll update those comments :p. Does it look alright aside from that?

dkanada: Yeah the conversion looks good.

PR #28, opened 2020-04-19
Add webpack

YouKnowBlom:
(Don't be alarmed by the large diff, most is from `yarn.lock`)

Adds webpack with build configs for production and development.


.gitignore
```diff
@@ -230,4 +230,9 @@ pip-log.txt
 #Mr Developer
 .mr.developer.cfg
 MediaBrowser.WebDashboard/dashboard-ui/.idea/
-.vscode
\ No newline at end of file
+.vscode
+
+#############
+## Node (NPM/Yarn)
+#############
```
dkanada, line 237: consistent comments would be nice :p the one above this has no border and no space after the hash character

webpack.config.js
```diff
@@ -0,0 +1,37 @@
...
+            PRODUCTION: JSON.stringify(isProduction)
+        })
+    );
+
+    if (!isProduction) {
+        config.devtool = "#inline-source-map";
+    }
+
+    return config;
+};
```
dkanada, line 37: EOL newlines are preferred I think

package.json
```diff
@@ -0,0 +1,25 @@
...
+  "dependencies": {
+    "clean-webpack-plugin": "^3.0.0",
+    "copy-webpack-plugin": "^5.1.1",
+    "file-loader": "^6.0.0",
+    "webpack": "4.42.1",
+    "webpack-cli": "^3.3.11"
+  },
+  "scripts": {
+    "devBuild": "webpack --config webpack.config.js --mode=development",
+    "prodBuild": "webpack --config webpack.config.js --mode=production"
```
dkanada, line 23: I think these could be improved. NPM usually has some standard commands like `prepare` that serve an actual function but it would also be nice to pick one flavor and label it `build` as the default.

.gitignore
YouKnowBlom: Yeah it's a mess :p. Tons of unrelated ignores too so I'll probably just clean all of it up

webpack.config.js
YouKnowBlom: Oh right, I forgot to specify that in vscode.

package.json
YouKnowBlom: That's a good idea. Maybe with a naming scheme similar to web?
`prepare` builds production.
`build:development`.
`build:production`.

dkanada: That sounds good to me.

YouKnowBlom:
There we go. Took one too many force pushes 😅

PR #31, opened 2020-04-28
Add eslint

YouKnowBlom:
Currently eslint is disabled for most files. Plan is to gradually fix them instead of shoving it all into this PR


.eslintrc.js
```diff
@@ -0,0 +1,44 @@
...
+    root: true,
+    env: {
+        node: true,
+        es6: true
+    },
+    extends: [
+        'eslint:recommended'
+    ],
+    parserOptions: {
+        ecmaVersion: 2015,
```
dkanada, line 11: might as well use 2020

YouKnowBlom: Sounds good. I wasn't sure if there was much support for es2020 yet but it seems like chrome supports a lot of the new stuff so we should be good👍

YouKnowBlom:
> Are these files transpiled for older browsers? I was planning on setting up a nightly but that would be a prerequisite since I'm sure a lot of clients are extremely old.

They are currently not transpiled. Afaik all cast devices run a recent version of chrome so I don't think that should be an issue, unless the user has somehow disabled automatic updates.

.eslintrc.js
MrTimscampi: You can probably throw Babel in there if you need another version in the browser :)

PR #45, opened 2020-07-28
fix bugs that were preventing successful casting

jasonmeisel:
There are three main fixes in here:

maincontroller.js
- the `playbackMgr` object was not being constructed with its constructor parameters.

playbackmanager.js
- many callback functions were using `this.` inside unbound anonymous functions, and thus was hitting undefined references. fixed by using arrow functions instead.
- `getNextPlaybackItemInfo` was failing because it uses `window.playlist`, which was not being assigned by `playbackManager`. Since removing `window.playlist` will require more refactoring, I just assign it before calling the function.

These fixes allow casting to work successfully, including media control using the Google Home app:
image

### DIFF
--- a/src/components/maincontroller.js
+++ b/src/components/maincontroller.js
@@ -34,7 +34,7 @@ window.mediaManager.addEventListener(cast.framework.events.category.CORE,
         console.log(event);
     });
 
-const playbackMgr = new playbackManager();
+const playbackMgr = new playbackManager(window.castReceiverContext, window.mediaManager);
 
 const playbackConfig = new cast.framework.PlaybackConfig();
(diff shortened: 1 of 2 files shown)


dkanada:
The changes should be available on the unstable version by the time you read this, feel free to make any other improvements since we'd love to tag a new stable release soon.

jasonmeisel:
> The changes should be available on the unstable version by the time you read this, feel free to make any other improvements since we'd love to tag a new stable release soon.

Thinking about it, some of these bugs would have been prevented if the codebase was using Typescript. I'd be happy to convert it, if you'd like.

If you haven't used Typescript before, it's *literally* just JS with type annotations. So the code changes would be fairly minor.

PR #46, opened 2020-07-28
use sender id rather than current time

dkanada:


neopc10:
I believe that this is a breaker.

PR #47, opened 2020-07-29
Convert codebase to Typescript

jasonmeisel:
Most of the changes are integrating Typescript or adding type annotations. However, some of the changes are bug fixes that were caught by the type checker - calls to undefined functions and references to undefined variables.

The most significant bug fix is in `jellyfinactions.ts`. The `factory` object contained functions that were calling other functions using `this.`. Since the functions were unbound, `this` was not `factory` but rather `undefined`. I fixed this by creating a `_Factory` class and instantiating it to initialize `factory`, which is a lot cleaner anyway (in my opinion).

I've tested the build locally, and it seems to work as well as the Javascript version.


src/app.ts
```diff
@@ -1,6 +1,29 @@
...
+        commandHandler,
+        deviceInfo,
+        reportEventType,
+        senderId,
+        subtitleAppearance,
+        repeatMode,
+    }
+}
+
+let senders = CastReceiverContext.getInstance().getSenders();
```
dkanada, line 26: This was reported as not working, was it fine when you tested? I applied it blindly since I don't own a cast device.

jasonmeisel: By "this" do you mean:
`let senders = cast.framework.CastReceiverContext.getSenders();`
Because it's been changed to:
`let senders = CastReceiverContext.getInstance().getSenders();`
😀

dkanada: Ah, I see that now :laughing: that would be the one!

YouKnowBlom: Oh I just realized you had fixed this here, after submitting #48.

YouKnowBlom:
Awesome work so far! I'll review it as soon as I get the chance.

`jellyfinactions` is indeed a bit of a mess. It's one of the files I had planned to rewrite eventually. I'll make an issue to write down some of my thoughts so we can discuss them, now that I'm not the only one working on this ;p.

Also, welcome!

ferferga:
Hello! @jasonmeisel 

Why this was closed? It was good!

PR #58, opened 2020-09-07
refactor changeStream nested promise chain

alexmaras:
Refactored the `changeStream` function to clean up a nested promise chain and make it more readable.
No logic changes have occurred here - the internal function has been made `async` in order to handle the promises in a more linear fashion and avoid the `new Promise`/`resolve()` pattern and still maintain order of operations

### DIFF
--- a/src/components/maincontroller.js
+++ b/src/components/maincontroller.js
@@ -337,48 +337,39 @@ export function changeStream(ticks, params) {
     var item = $scope.item;
     var mediaType = item.MediaType;
 
-    // TODO untangle this shitty callback mess
-    getMaxBitrate(mediaType).then(function (maxBitrate) {
-        var deviceProfile = getDeviceProfile(maxBitrate);
+    getMaxBitrate(mediaType).then(async (maxBitrate) => {
+        const deviceProfile = getDeviceProfile(maxBitrate);
+        const audioStreamIndex = params.AudioStreamIndex == null ? $scope.audioStreamIndex : params.AudioStreamIndex;
+        const subtitleStreamIndex = params.SubtitleStreamIndex == null ? $scope.subtitleStreamIndex : params.SubtitleStreamIndex;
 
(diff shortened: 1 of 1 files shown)


YouKnowBlom:
Awesome! If it is one thing this client needs it's definitely refactoring. Will take a look ASAP.

alexmaras:
Great, thanks! I noticed https://github.com/jellyfin/jellyfin-chromecast/pull/59 working towards some similar improvements - I'd be happy to help out with an effort to convert to some of those newer constructs (`async`/`await`, `const` etc)

Is there documentation on how best to test this codebase locally?

YouKnowBlom:
> I'd be happy to help out with an effort to convert to some of those newer constructs (async/await, const etc)

Please do, that would be very much appreciated! :)

> Is there documentation on how best to test this codebase locally?

There isn't currently a good way to test without a cast device unfortunately. I've looked into it a little bit and there seems to be ways to emulate a device with [this](https://github.com/ajhsu/chromecast-device-emulator) and test that way, but I haven't had the time to look into it further.

SteveDinn:
I'd like to help test this as well. I'm anxious to have a better chrommecast client, but I'm not sure how (or where) to apply any updates.

neopc10:
> I'd like to help test this as well. I'm anxious to have a better chrommecast client, but I'm not sure how (or where) to apply any updates.

I think maincontroller.js is in need of [Queueing ](https://developers.google.com/cast/docs/web_receiver/core_features)support for the web receiver and then [Cast Connect](https://developers.google.com/cast/docs/android_tv_receiver) receiver support for the Android TV repo in preparation for the new devices due out.

The Android 2.0.0 repo is in need of a new [Android sender](https://developers.google.com/cast/docs/android_sender/intent_to_join), as I believe it is currently missing.  If jellyfin-web's Google Cast ever gets separated away, I'm sure the Chrome sender would benefit from a one-over.

YouKnowBlom (approved):
Sorry for the delay, #60 blocked me when I got the time to test. Works perfectly from my testing, thanks for the PR :)

PR #65, opened 2020-09-20
Fix playlist index always set to 0

### LINKED ISSUE #54: Cannot play specific track in album when chromecasting audio
**Describe the bug**
When a cast session is open to any chromecast-compatible device (actual chromecast or Google Home device), selecting any given track in an album will always start play at the first track of the album... and it will play sequentially. The playlist only contains the current track, so the skip buttons are nonfunctional.

**System (please complete the following information):**
 - OS: Ubuntu 18.04
 - Virtualization: Docker
 - Clients: Browser + ChromeCast
 - Browser: Chromium Version 83.0.4103.97 (Official Build) snap (64-bit)
 - Jellyfin Version: 10.5.5
 - Playback: ...chromecast?
 - Installed Plugins:
   - Anime: 5.0.0.0
   - AudioDB: 10.5.5.0
   - MusicBrainz: 10.5.5.0
   - Open Subtitles: 7.0.0.0
 - Reverse Proxy: nginx
 - Base URL: jellyfin.MYURL.com
 - Networking: erm... whatever the docker container uses.
 - Storage: SMB connection to NAS

**To Reproduce**
1. Open an album in a music library consisting of more than one track
2. Start a chromecast connection to any chromecast-capable device
3. Click on a track that is not the first track in the album.

**Expected behavior**
The track that you clicked on should start playing through the device that you're casting to. 

**Logs**
None.

**Screenshots**

**Additional context**

YouKnowBlom:
Previously playlist index was always set to 0 before playback started, which caused playlists (and albums) to start from the beginning.

Fixes #54

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -52,30 +52,31 @@ export function getReportingParams($scope) {
 
 export function getNextPlaybackItemInfo() {
 
-    var playlist = window.playlist;
+    var playlist = window.playbackManager.activePlaylist;
 
     if (!playlist) {
         return null;
     }
 
+    let activeIndex = window.playbackManager.activePlaylistIndex;
     var newIndex;
 
-    if (window.currentPlaylistIndex == -1) {
+    if (activeIndex == -1) {
         newIndex = 0;
     } else {
         switch (window.repeatMode) {
 
             case 'RepeatOne':
-                newIndex = window.currentPlaylistIndex;
+                newIndex = activeIndex;
                 break;
             case 'RepeatAll':
-                newIndex = window.currentPlaylistIndex + 1;
-                if (newIndex >= window.playlist.length) {
+                newIndex = activeIndex + 1;
+                if (newIndex >= playlist.length) {
                     newIndex = 0;
                 }
                 break;
             default:
-                newIndex = window.currentPlaylistIndex + 1;
+                newIndex = activeIndex;
                 break;
         }
     }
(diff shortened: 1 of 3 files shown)


src/helpers.js
```diff
@@ -52,30 +52,31 @@ export function getReportingParams($scope) {
...
-                newIndex = window.currentPlaylistIndex + 1;
-                if (newIndex >= window.playlist.length) {
+                newIndex = activeIndex + 1;
+                if (newIndex >= playlist.length) {
                     newIndex = 0;
                 }
                 break;
             default:
-                newIndex = window.currentPlaylistIndex + 1;
+                newIndex = activeIndex;
```
YouKnowBlom, line 79: I didn't notice any anomalies from removing the `+ 1` during my fairly extensive testing. I believe this was done to prevent the index from being -1 but those cases seems to be taken care of by this if statement above:

```js
if (activeIndex == -1) {
    newIndex = 0;
} else {
```

dkanada: This looks like it was done to increment the song during playback, otherwise it will always repeat the same song. The code for that could be in another location though, I haven't checked.

YouKnowBlom: That would make a lot of sense. All the queue stuff that was implemented has since stopped working (if it did post-CAF migration, can't quite remember) so I didn't test that scenario

YouKnowBlom:
I'm working on a full rewrite to use the CAF queue system and since this is a bit of hack I'll close it

PR #69, opened 2020-10-31
Add typescript and eslint-typescript, convert app.js

YouKnowBlom:
Add and configure typescript, cast types and eslint-typescript
Start migration by converting app.js

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,8 +33,8 @@ module.exports = {
         'semi': ["error"],
         'space-before-blocks': ["error"],
         // Disable these until we have converted the project to TS
-        '@typescript-eslint/explicit-module-boundary-types': "off"
-        
+        '@typescript-eslint/explicit-module-boundary-types': "off",
+        '@typescript-eslint/no-explicit-any': "off"
     },
     overrides: [{
         files: ['./src/**/*.js', './src/**/*.ts'],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -31,22 +31,29 @@ module.exports = {
         'no-trailing-spaces': ["error"],
         'one-var': ["error", "never"],
         'semi': ["error"],
-        'space-before-blocks': ["error"],
-        // Disable these until we have converted the project to TS
-        '@typescript-eslint/explicit-module-boundary-types': "off",
-        '@typescript-eslint/no-explicit-any': "off"
+        'space-before-blocks': ["error"]
     },
-    overrides: [{
-        files: ['./src/**/*.js', './src/**/*.ts'],
-        env: {
-            node: false,
-            browser: true,
-            es6: true
+    overrides: [
+        {
+            files: ['./src/**/*.js', './src/**/*.ts'],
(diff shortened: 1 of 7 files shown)


.eslintrc.js
```diff
@@ -26,10 +31,13 @@ module.exports = {
         'no-trailing-spaces': ["error"],
         'one-var': ["error", "never"],
         'semi': ["error"],
-        'space-before-blocks': ["error"]
+        'space-before-blocks': ["error"],
+        // Disable these until we have converted the project to TS
```
MrTimscampi, line 35: You could do overrides for TypeScript in order to allow the TS rules only for `.ts` files.

src/app.ts
```diff
@@ -1,7 +1,24 @@
 import "./components/maincontroller";
 
-let senders = cast.framework.CastReceiverContext.getInstance().getSenders();
-let id = senders.length !== 0 && senders[0].id ? senders[0].id : new Date().getTime();
+declare global {
```
MrTimscampi, line 3: You can usually create a `d.ts` file for this an put it anywhere in the project.

It'll define the type in the entire project.

src/app.ts
```diff
@@ -1,7 +1,24 @@
...
+    interface Window {
+        deviceInfo: deviceInfo;
+        mediaElement: HTMLElement;
+        playlist: Array<any>;
+        currentPlaylistIndex: number;
+        repeatMode: "RepeatOne" | "RepeatAll" | "RepeatNone";
+    }
+}
+
+interface deviceInfo {
```
MrTimscampi, line 14: Best practice is generally to use PascalCase for interface names. It avoid confusing them with variables.

src/app.ts
```diff
@@ -1,7 +1,24 @@
...
-let senders = cast.framework.CastReceiverContext.getInstance().getSenders();
-let id = senders.length !== 0 && senders[0].id ? senders[0].id : new Date().getTime();
+declare global {
+    let PRODUCTION: boolean;
+    interface Window {
+        deviceInfo: deviceInfo;
+        mediaElement: HTMLElement;
+        playlist: Array<any>;
+        currentPlaylistIndex: number;
+        repeatMode: "RepeatOne" | "RepeatAll" | "RepeatNone";
```
MrTimscampi, line 10: It could be interesting down the line to use an enum for this.

You'd essentially be able to use `RepeatMode.RepeatOne` or similar everywhere.

tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+    "compilerOptions": {
```
MrTimscampi, line 2: You should probably also add:
```json
{
  "strict": true,
}
```
It'll enable a bunch of crap for you to ensure stricter typing, which is usually a good idea.

MrTimscampi (commented):
Looks good so far :+1:

src/app.ts
YouKnowBlom: Yeah that's a good idea

PR #72, opened 2020-11-18
Add axios

YouKnowBlom:
Adds axios as a dependency
Adds a basic credential manager for use later
Adds api.ts generated from openapi.json (2020-11-15)
Copied the `.gitattributes` from jellyfin-vue


Camc314:
Since it's building, I think you've already fixed this. But when I produced the API for jf-vue, on the latest stable openapi generator (v4.3.1), lots of typescript errors are produced. Instead use the 5.0.x beta.

See [here](https://github.com/jellyfin/jellyfin/pull/4070#discussion_r498851097)

YouKnowBlom:
Yeah I had that issue too but I remember seeing you talk about it in matrix so I backtracked to a fix :p

PR #78, opened 2020-11-22
Append date and revision to package.json version field

YouKnowBlom:
Append `Build.BuildNumber` to version field in package.json
Use package.json version field in deviceInfo.versionNumber


EraYaN (approved):
Looks good

PR #79, opened 2020-11-22
Migrate jellyfinactions

YouKnowBlom:
Migrates jellyfinactions to typescript
Updates all references to jellyfinactions

Refactor to come :)


src/types/global.d.ts
```diff
@@ -1,17 +1,27 @@
...
+}
+
+export interface GlobalScope {
+    [key: string]: any
+}
+
 declare global {
     export const PRODUCTION: boolean;
     export const RECEIVERVERSION: string;
+    export let $scope: GlobalScope;
```
ferferga, line 16:
```suggestion
    export const $scope: GlobalScope;
```

Why not const?

YouKnowBlom: Ah my bad, I thought it was already const. I must've mixed it up with something else :p

YouKnowBlom:
Dang, commiting via the github app just puts a generic commit message. I'll update that later today

ferferga (approved):

PR #82, opened 2020-11-28
Add jest and config

YouKnowBlom:
Add jest and basic config.
Add test task to github actions.


package.json
```diff
@@ -18,10 +18,12 @@
     "prepare": "webpack --config webpack.config.js --mode=production",
     "build:production": "webpack --config webpack.config.js --mode=production",
     "build:development": "webpack --config webpack.config.js --mode=development",
-    "lint": "eslint ./src/"
+    "lint": "eslint ./src/",
+    "test": "jest"
```
simonmarklar, line 22: might want to add a `pretest` script as well that runs the lint script. might as well catch those errors early :)

also, if eslint takes a while you can also add ` --fix --cache` to the end of the lint script above

YouKnowBlom: Went ahead and added the pretest script :+1:

.github/workflows/main.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+on:
+    push:
+        branches:
+            - main
+            - master
+    pull_request:
+        branches:
+            - main
+            - master
```
nielsvanvelzen, line 11: This repository uses the branch "master" so you can drop the "main" from those lists

.github/workflows/main.yaml
```diff
@@ -0,0 +1,48 @@
+name: Main
```
nielsvanvelzen, line 1: Main is an ambiguous name. I'd call it something like "test" or "build"

.github/workflows/main.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+    jest:
+        name: Jest
+        runs-on: ${{ matrix.os }}
+
+        strategy:
+            matrix:
+                os: [ubuntu-latest]
+                node: [14]
```
nielsvanvelzen, line 21: This matrix is useless as it only has a single value. In this case I'd say it's easier to hardcode the values or add more entries (like testing both node latest and LTS). (the OS should always stay the same though, since the project doesn't use platform specific features)

.github/workflows/main.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+        steps:
+            - name: Checkout
+              uses: actions/checkout@master
```
nielsvanvelzen, line 25: Use versioned actions:

```suggestion
              uses: actions/checkout@2
```

.github/workflows/main.yaml
```diff
@@ -0,0 +1,48 @@
...
+            - name: Cache node_modules
+              uses: actions/cache@v2
+              id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+              with:
+                  path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+                  key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+                  restore-keys: |
+                      ${{ runner.os }}-yarn-
+            - name: Install dependencies
+              run: yarn
```
nielsvanvelzen, line 45:
```suggestion
              run: yarn install --frozen-lockfile
```

this argument is made for CI, see the docs for more info

.github/workflows/main.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+            - name: Run tests
+              run: yarn test
```
nielsvanvelzen, line 48:
```suggestion
              run: yarn run test
```

Using explicit naming for the commands will help in case yarn ever adds a custom command called "test".

nielsvanvelzen (changes requested):
I reviewed the workflow only

nielsvanvelzen (approved):

PR #85, opened 2020-11-29
Add development section to CONTRIBUTING.md

YouKnowBlom:


CONTRIBUTING.md
```diff
@@ -2,6 +2,15 @@
 
 ## Development
 
+1. Register a new [application](https://developers.google.com/cast/docs/registration).
```
simonmarklar, line 5: Could you expand on this step? eg note that its a "Custom Receiver", and does the name matter? does it have to match a property in the source or can it be anything?

CONTRIBUTING.md
```diff
@@ -2,6 +2,15 @@
...
+2. Set up a local copy of [jellyfin-web](https://github.com/jellyfin/jellyfin-web).
+3. Change `applicationStable` and `applicationUnstable` in `jellyfin-web/src/plugins/chromecastPlayer/plugin.js` to your own application ID.
+4. Clone repo and run `yarn run`. This will install all dependencies and build a production build by default.
+5. Make changes and build with `yarn build:development`. `build:development` includes sourcemaps that'll make it easier to debug.
+6. Before pushing your changes, make sure to run `yarn test` and `yarn lint`.
+
+> NOTE: It is recommended to symlink the dist folder to the location on your web server hosting the files. That way you can refresh the cast receiver via the Chrome Remote Debugger and see your changes without having to manually copy after each build.
```
simonmarklar, line 12: could you say what the target location of the symlink would be?

YouKnowBlom: This will vary heavily depending on web server but I suppose it could be phrased a little better

CONTRIBUTING.md
YouKnowBlom: Those are some good points 👍

CONTRIBUTING.md
simonmarklar: ah, i thought you meant that the chromecast source needs to be symlinked to a folder in the jellyfin-web (which i just assumed would be set to be the webserver root). perhaps just clarify what `dist` is being symlinked?

dkanada:
@YouKnowBlom do you plan on moving this pull request out of draft mode?

YouKnowBlom:
Yeah that's the plan! I have some time next week so I'll finish up and go through most of the PRs then. @dkanada

dkanada (approved):

PR #86, opened 2020-11-30
make compatible with chromium 83

hawken93:
Chromium likes to pass event data json encoded, so I made a check to deal with that.
Also 2 BOM removals

### DIFF
--- a/src/components/jellyfinActions.ts
+++ b/src/components/jellyfinActions.ts
@@ -1,4 +1,4 @@
-﻿import { ajax } from './fetchhelper';
+import { ajax } from './fetchhelper';
 
 import {
     getUrl,
--- a/src/components/maincontroller.js
+++ b/src/components/maincontroller.js
@@ -488,6 +488,13 @@ window.castReceiverContext.addCustomMessageListener(
 
         var data = evt.data;
 
+        // Apparently chromium likes to pass it as json, not as object.
+        // chrome on android works fine
+        if (typeof data == 'string' || data instanceof String) {
+            console.log('Event data is a string.. Chromium detected..');
+            data = JSON.parse(data);
+        }
+
         data.options = data.options || {};
         data.options.senderId = evt.senderId;
         // TODO set it somewhere better perhaps
--- a/src/components/maincontroller.js
+++ b/src/components/maincontroller.js
@@ -1,4 +1,4 @@
-﻿/* eslint-disable */
+/* eslint-disable */
 
 import {
     reportPlaybackProgress,


YouKnowBlom:
I thought all images were included by `html-webpack-plugin` but I see now that I missed a few, good catch :p. I think we should try to make use of the `css-loader` plugin as that parses `url()`s by default

https://webpack.js.org/loaders/css-loader/#url

hawken93:
I've messed around some more and made it work the way you wanted I think. But please check for yourself too. I also noticed favicon didn't have anything to help it. Keeping in mind that it needs to retain its filename I just copied it over. Without the copy plugin, webpack seemed to ignore it.

src/components/jellyfinActions.ts
```diff
@@ -1,4 +1,4 @@
-﻿import { ajax } from "./fetchhelper";
+import { ajax } from "./fetchhelper";
```
hawken93, line 1: remove BOM

src/components/maincontroller.js
```diff
@@ -1,4 +1,4 @@
-﻿/* eslint-disable */
+/* eslint-disable */
```
hawken93, line 1: remove BOM

src/css/jellyfin.css
```diff
@@ -208,11 +208,11 @@ body {
 }
 
 .fresh {
-    background-image: url(../img/fresh.svg);
+    background-image: url('../img/fresh.svg');
```
hawken93, line 211: husky hook wanted single quotes so that's the only change to the css that's left. Now the path gets rewritten instead so that's nice.

src/index.html
```diff
@@ -3,6 +3,7 @@
 <head>
     <meta charset="utf-8">
     <title>Jellyfin</title>
+    <link rel="shortcut icon" href="favicon.ico">
```
hawken93, line 6: Failed attempt at getting webpack to include it, but I figured it should stay in

webpack.config.js
```diff
@@ -21,13 +22,17 @@ let config = {
...
             template: 'index.html',
             hash: false
-        })
+        }),
+        new CopyPlugin([{ from: 'favicon.ico', to: 'favicon.ico' }])
     ],
     module: {
         rules: [
-            { test: /\.html$/, loader: "html-loader"},
-            { test: /\.(svg|png|jpe?g|gif|eot|woff|tff)$/i, loader: "url-loader" },
```
hawken93, line 29: Did I break anything by changing this to file-loader?

hawken93:
I've added some comments to make it easy to understand my intention with every change

hawken93:
Looks like it messes up production mode..

hawken93:
I've reverted back to my copy approach after css-loader introduced more mess than I was able to fix. I also discovered fonts were being left out, so I've fixed that too

hawken93:
If anyone can make it work with css-loader, please open a separate PR and merge that, then I'll make corresponding changes here. :)

hawken93:
Recommend solving the webpack stuff here:
https://github.com/jellyfin/jellyfin-chromecast/pull/88

I'll make this PR only about the chromium workaround following merging that one :)

YouKnowBlom:
Converting this to a draft until #88 is ready :p

hawken93:
rebased on top of webpack-fix PR, this PR will get much cleaner after that one is merged :)

YouKnowBlom (approved):

hawken93:
Updated title and description to match the content of the pr

PR #89, opened 2020-12-05
Various fixes to make the chromecast behave better

hawken93:
I still have problems that it doesn't select or change subs properly. AC3 still not working, see https://issuetracker.google.com/issues/174837344

### DIFF
--- a/src/components/maincontroller.js
+++ b/src/components/maincontroller.js
@@ -34,10 +34,7 @@ import { JellyfinApi } from './jellyfinApi';
 window.castReceiverContext = cast.framework.CastReceiverContext.getInstance();
 window.mediaManager = window.castReceiverContext.getPlayerManager();
 
-const playbackMgr = new playbackManager(
-    window.castReceiverContext,
-    window.mediaManager
-);
+const playbackMgr = new playbackManager(window.mediaManager);
 
 var init = function () {
     resetPlaybackScope($scope);
@@ -276,7 +273,6 @@ export function processMessage(data) {
 
     if (!cmdHandler) {
         window.commandHandler = new commandHandler(
-            window.castReceiverContext,
             window.mediaManager,
             playbackMgr
         );
--- a/src/components/maincontroller.js
+++ b/src/components/maincontroller.js
@@ -33,23 +33,11 @@ import { JellyfinApi } from './jellyfinApi';
 
 window.castReceiverContext = cast.framework.CastReceiverContext.getInstance();
 window.mediaManager = window.castReceiverContext.getPlayerManager();
-window.mediaManager.addEventListener(
-    cast.framework.events.category.CORE,
-    (event) => {
-        console.log('Core event: ' + event.type);
-        console.log(event);
-    }
-);
 
 const playbackMgr = new playbackManager(
     window.castReceiverContext,
     window.mediaManager
 );
-
-const playbackConfig = new cast.framework.PlaybackConfig();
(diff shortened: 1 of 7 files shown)


hawken93:
This one is on top of my other PR.
The main changes are:
- fix the subtitle track if-block logic error
- pick different API urls than the /emby/ ones and lowercased the directstream /Videos/ endpoint
- cleaned up the call into deviceProfile, made arguments mandatory. Moved ac-3 check into codecSupportHelper.
- fixed the bug that causes it to always assume 6 channels available (`any positive number` ? 6 : 2) == 6
- prepped for AC-3 and E-AC-3 support, but made workarounds to disable them for now.
- Made a PlayNow command handler, shouldn't rely on the default handler. Increased visibility of default handler usage.
- Prolong or disable dev mode timeout (I think I'll completely disable it)
- removed some unused constructor arguments
- moved some context stuff into maincontroller to try making sure that the receiver context isn't instantiated everywhere, goal is to pass one instance around.
- Simplified getMaxBitrate and made a sane fallback for it; set StaticBitrate as well because the server seems to assume that the available bitrate is always too low

I still have issues with:
- Instability, it likes to quit on you
- Subtitle change during playback crashes it
- That API queuing is not being used
- AC-3 still not supported; pending my opened issue over at google

src/helpers.js
```diff
@@ -421,7 +421,7 @@ export function createStreamInfo(item, mediaSource, startPosition) {
         let subStreamCodec = subtitleStream.Codec.toLowerCase();
         if (subStreamCodec !== 'vtt' &&
             subStreamCodec !== 'webvtt' &&
-            (subStreamCodec === 'srt' && subtitleStream.DeliveryMethod !== "External")) {
+            !(subStreamCodec === 'srt' && subtitleStream.DeliveryMethod !== "External")) {
```
YouKnowBlom, line 424: Have you tested this with external srts? I think this change is breaking that

YouKnowBlom: Wait no, I'm being stupid here hehe

hawken93: I can invert this logic so it'll be human readable

hawken93:
text style change --> see commit message, I decided to write it out in there.

hawken93:
The subtitle change doesn't always work, I was thinking to push settings when the app starts instead, and maybe again when transferring the settings if there is a change to them.

YouKnowBlom:
Oh yeah we've been talking about doing that a while back and I definitely agree 👍

src/components/maincontroller.js
```diff
@@ -694,14 +678,30 @@ export function setTextTrack(index) {
         if (subtitleTrack) {
             textTracksManager.setActiveByIds([subtitleTrack.trackId]);
             var subtitleAppearance = window.subtitleAppearance;
+
+            // https://developers.google.com/cast/docs/reference/chrome/chrome.cast.media.TextTrackStyle
+            var textTrackStyle = new cast.framework.messages.TextTrackStyle();
```
hawken93, line 683: this stuff doesn't work right now because I don't think it necessarily gets called

hawken93: removed from the PR

hawken93:
Subtitle style fix removed from this PR, I need to rethink this part..

hawken93:
I added a debug statement because the server address seems to get nulled sometimes.
It would be nice to see if that happens through the setter or happens due to something else.

YouKnowBlom (approved):

PR #91, opened 2020-12-06
Filtered set of the various fixes PR

hawken93:
The ones I think are actually useful. Messing around with logging and dev mode idle timeout was too messy for this round.

* Fix the subtitle check
* Remove emby references in URLs, lowercase /Videos/ to /videos/ in the directstream case
    * (I wish we could agree on a best practice of using the lowercase variants. The server seems case insensitive.)
* Add PlayNow command handler, increase verbosity of default handler invocations
* Start supporting AC-3 and E-AC-3, but disable it in codecSupportHelper because CAF is broken (it seems)
    * Clean up deviceProfile call, make arguments mandatory. Move the ac-3 check to codecSupportHelper.
    * Now everyone gets 2ch aac... Can someone show me 6ch aac working? Does that work on any chromecast?
* Clean up the max bitrate code because it was transcoding due to exceeded bandwidth when that was false.

### DIFF
--- a/src/components/deviceprofileBuilder.ts
+++ b/src/components/deviceprofileBuilder.ts
@@ -267,7 +267,7 @@ function getTranscodingProfiles(): Array<TranscodingProfile> {
     const TranscodingProfiles: Array<TranscodingProfile> = [];
 
     const hlsAudioCodecs = getSupportedHLSAudioCodecs();
-    const audioChannels: number = hasSurroundSupport() ? 6 : 2;
+    const audioChannels: number = hasSurroundSupport(currentDeviceId) ? 6 : 2;
 
     if (profileOptions.enableHls !== false) {
         TranscodingProfiles.push({
--- a/src/components/deviceprofileBuilder.ts
+++ b/src/components/deviceprofileBuilder.ts
@@ -16,6 +16,7 @@ import { ProfileConditionValue } from '../api/generated/models/profile-condition
 import { deviceIds, getActiveDeviceId } from './castDevices';
 
 import {
+    hasSurroundSupport,
     hasTextTrackSupport,
     hasVP8Support,
     hasVP9Support,
@@ -33,7 +34,6 @@ import {
 } from './codecSupportHelper';
 
 interface ProfileOptions {
-    audioChannels: number;
     enableHls: boolean;
     bitrateSetting?: number;
 }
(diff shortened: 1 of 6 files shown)


hawken93:
Investigating if this breaks external srt support

hawken93:
It did but now it doesn't :)

Instead of the complicated clause, I just check if the URL is defined or not. If it's undefined, that means the server doesn't have an externally loaded subtitle available.

This URL should always be playable because of the profiles that have been set.

If the server adds more things to the list that the server can transcode to vtt, we will be able to get those improvements without code changes.

src/helpers.js
```diff
@@ -469,7 +467,8 @@ export function createStreamInfo(item, mediaSource, startPosition) {
         console.log('Subtitle url: ' + info.subtitleStreamUrl);
     });
 
-    if (subtitleTracks) {
+    if (subtitleTracks && subtitleTracks.length) {
```
hawken93, line 470: I made this change trying to get the tracks undefined if there were none, but I think I didn't understand how things worked at the time, so this might nt actually do anything useful.
(for reference: `if (subtitleTracks && subtitleTracks.length)`)

src/helpers.js
```diff
@@ -469,9 +467,7 @@ export function createStreamInfo(item, mediaSource, startPosition) {
         console.log('Subtitle url: ' + info.subtitleStreamUrl);
     });
 
-    if (subtitleTracks) {
-        info.tracks = subtitleTracks;
-    }
+    info.tracks = subtitleTracks;
```
hawken93, line 470: My conclusion is that `tracks` refer to text tracks and that an empty array here doesn't hurt.

The previous if-check always returns true, so this should have no impact.

https://developers.google.com/cast/docs/reference/web_receiver/cast.framework.messages.MediaInformation

src/components/deviceprofileBuilder.ts
```diff
@@ -362,17 +363,18 @@ function getSubtitleProfiles(): Array<SubtitleProfile> {
...
     // MaxStaticBitrate seems to be for offline sync only
     const profile: DeviceProfile = {
-        MaxStreamingBitrate: bitrateSetting,
-        MaxStaticBitrate: 0,
-        MusicStreamingTranscodingBitrate: Math.min(bitrateSetting, 192000)
+        MaxStreamingBitrate: options.bitrateSetting,
+        MaxStaticBitrate: options.bitrateSetting,
+        MusicStreamingTranscodingBitrate: Math.min(
+            options.bitrateSetting,
+            192000
```
hawken93, line 376: is there a good reason for this 192k cap?

YouKnowBlom: 1. Nah that's probably a good idea to change
2. I believe MaxStaticBitrate is leftovers from emby for syncing content but I could be wrong

YouKnowBlom (approved):

gabri94:
Has this been included in a release?

YouKnowBlom:
It has not yet made it to stable but unstable has it @gabri94

gabri94:
Unfortunately the unstable branch crashes every time i play a video

PR #92, opened 2020-12-12
JellyfinApi static class for server information

hawken93:
All ajax calls and url constructors have been refactored to be handled by JellyfinApi

### DIFF
--- a/src/components/jellyfinApi.ts
+++ b/src/components/jellyfinApi.ts
@@ -1,12 +1,6 @@
-// this class will take over for any ajax call.
-
 import { ajax } from './fetchhelper';
+import { Dictionary } from '../types/global';
 
-interface Dictionary<T> {
-    [Key: string]: T;
-}
-
-// DocumentManager should never ha
 export abstract class JellyfinApi {
     // userId that we are connecting as currently
     public static userId: string | null = null;
--- a/src/components/jellyfinApi.ts
+++ b/src/components/jellyfinApi.ts
@@ -1,0 +1,81 @@
+// this class will take over for any ajax call.
+
+import { ajax } from './fetchhelper';
+
+interface Dictionary<T> {
+    [Key: string]: T;
+}
+
+// DocumentManager should never ha
+export abstract class JellyfinApi {
+    // userId that we are connecting as currently
+    public static userId: string | null = null;
+
+    // Security token to prove authentication
+    public static accessToken: string | null = null;
+
+    // Address of server
+    public static serverAddress: string | null = null;
+
+    public static setServerInfo(
+        userId: string,
+        accessToken: string,
+        serverAddress: string
+    ) {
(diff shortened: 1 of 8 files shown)


hawken93:
Also reenabled eslint for the files and fixed linting errors.

src/components/jellyfinApi.ts
```diff
@@ -0,0 +1,81 @@
+// this class will take over for any ajax call.
+
+import { ajax } from './fetchhelper';
+
+interface Dictionary<T> {
+    [Key: string]: T;
+}
```
hawken93, line 7: Maybe this could be moved somewhere else, or maybe there is another type better fit for a dict of string->string?

src/components/maincontroller.js
```diff
@@ -184,15 +179,10 @@ mgr.addEventListener(cast.framework.events.EventType.ENDED, function () {
     reportPlaybackStopped($scope, getReportingParams($scope));
     init();
 
-    if (!playNextItem()) {
+    if (!playbackMgr.playNextItem()) {
```
hawken93, line 182: This got caught by the linter, playNextItem was not defined, so this would break playing the next item in the queue I think.

src/components/maincontroller.js
```diff
@@ -204,9 +194,7 @@ window.castReceiverContext.addEventListener(
                 JSON.stringify(event.data)
         );
 
-        if ($scope.userId != null) {
-            reportEvent('volumechange', true);
-        }
+        reportEvent('volumechange', true);
```
hawken93, line 197: I guess we could do `if (JellyfinApi.userId)` - but I removed so many "do we have a connection" checks. I'm assuming someone would connect before these calls would happen.

src/components/maincontroller.js
```diff
@@ -378,15 +373,15 @@ export function setSubtitleStreamIndex($scope, index, serverAddress) {
     ) {
         var textStreamUrl = subtitleStream.IsExternalUrl
             ? subtitleStream.DeliveryUrl
-            : getUrl(serverAddress, subtitleStream.DeliveryUrl);
+            : JellyfinApi.createUrl(subtitleStream.DeliveryUrl);
```
hawken93, line 376: I haven't checked if we end up with a double slash here.

hawken93: Fixed

src/components/jellyfinApi.ts
YouKnowBlom: We could move this to `.d.ts` file in `src/types`

YouKnowBlom (changes requested):

src/components/jellyfinApi.ts
hawken93: did :)

YouKnowBlom (approved):

PR #93, opened 2020-12-12
DocumentManager

hawken93:
Making DocumentManager and taking the functions that deal with DOM in there.

Also fixed:
* Image preloading
* less DOM work for ccast audio
* Size of poster image in details page
* linting


src/components/documentManager.ts
```diff
@@ -0,0 +1,533 @@
+import { BaseItemDto } from '../api/generated/models/base-item-dto';
+
+import { JellyfinApi } from './jellyfinApi';
+
+import { parseISO8601Date } from '../helpers';
+
+// Duration between each backdrop switch
+const backdropPeriodMs: number = 30000;
+
+// DocumentManager should never ha
```
YouKnowBlom, line 10: Ha what?

hawken93: Probably an unfinished justification for having it as a static class

src/components/documentManager.ts
```diff
@@ -0,0 +1,532 @@
...
+    //playedPercentage: number | null;
+
+    // TODO: Static constructor if that's needed.
+    // For now, this is not in use and won't be run.
+
+    // tool for showItem
+    private static getPrimaryImageUrl(item: BaseItemDto): string | null {
+        if (item.AlbumPrimaryImageTag)
+            return JellyfinApi.createUrl(
+                'Items/' +
```
simonmarklar, line 28: If you didnt know, there's a relatively new feature in JS you can use to make this (imho) more legible than string concatenation - "Template Literals" - so you can ditch the string concat and the return statement would become

```
return JellyfinApi.createUrl(`Items/${item.AlbumId}/Images/Primary?tag=${item.AlbumPrimaryImageTag}`);
```

src/components/documentManager.ts
```diff
@@ -0,0 +1,532 @@
...
+        }
+
+        this.setDetailImage(detailImageUrl);
+    }
+
+    // It has some overloadings
+    public static showItemId(itemId: string): void {
+        const url = JellyfinApi.createUserUrl('Items/' + itemId);
+
+        JellyfinApi.authAjax(url, 'GET').then(function (item: BaseItemDto) {
```
simonmarklar, line 130: might not be a big deal in this case but its probably worthwhile returning this promise so whatever calls this function can wait for it to finish and be sure that the dom changes have been applied

src/components/documentManager.ts
```diff
@@ -0,0 +1,532 @@
...
+            DocumentManager.showItem(item);
+        });
+    }
+
+    private static getRatingHtml(item: BaseItemDto): string {
+        var html: string = '';
+        if (item.CommunityRating != null) {
+            html +=
```
simonmarklar, line 138: highly recommend using template literals for html, its pretty much the main reason they exist :)

src/components/documentManager.ts
```diff
@@ -0,0 +1,532 @@
+import { BaseItemDto } from '../api/generated/models/base-item-dto';
+
+import { JellyfinApi } from './jellyfinApi';
+
+import { parseISO8601Date } from '../helpers';
+
+// Duration between each backdrop switch in ms
+const backdropPeriodMs: number = 30000;
+
+export abstract class DocumentManager {
```
simonmarklar, line 10: Sorry if I am missing something, but what value does making a class with all static private/public methods/props bring vs. a typescript module which exports the public methods (but as functions)? 

When I first started reading this I assumed because it was an abstract class there would be another class which implements it. Instead its pretty much doing the job of a module but with an extra cognitive load.

hawken93: I've tried to write a long and complicated answer, but to make it easy: It's functionally the same but I find it more intuitive when it's organized into an object. For me, it decreases cognitive load rather than increase it.

In this case, before the calls into DocumentManager, you can see in the other files just by looking at the calls that the functions are inside this class. By making sure all html manipulations are in this class, you'll see by the "DocumentManager" prefix that this call does a task related to html manipulation.
I know that you can kind of mess with the imports to get something of the same effect, but I just like this style better.

hawken93:
Thanks for the review! I'll get back to it :)

src/components/documentManager.ts
hawken93: Also I should probably move the adjacent const into the class just to avoid mixing OOP and non-OOP

hawken93: But mostly this is personal preference

simonmarklar: yeah its functionally the same, I guess my main problem with it is that there is no concept of an abstract class in javascript. Even classes are really just sugar on top of `Function`. But that is my personal preference and there's no technical reason to choose one way over the other. I was just curious as to why you chose to use a class, Perhaps there was something I wasn't aware of that i could learn from

I do like that you've brought everything DOM related into one place, brings more structure to the project :)

hawken93: I was thinking about my decision to leave the backdropPeriodMs variable outside the class.

I would ideally make this something configurable. So that the user connecting to this chromecast can set this property to something that suits them. Or disable it completely.

We also have some user over at reddit who would prefer to switch the whole idle background image. In this case it would make sense to be able to pull this from the server.

I'll refactor this to a setter/getter style thing so it can be done easily later on :)

hawken93:
@simonmarklar I've done changes you suggested and I think I'm ready for another review :)

src/components/documentManager.ts
```diff
@@ -0,0 +1,512 @@
...
+        if (this.backdropPeriodMs === null) return;
+
+        this.setRandomUserBackdrop();
+
+        this.backdropTimer = <any>setInterval(function () {
+            DocumentManager.setRandomUserBackdrop();
+        }, this.backdropPeriodMs);
+    }
+
+    // null disables backdrops
```
hawken93, line 233: TODO here is to clear the backdrop if it gets set to null. Relatively minor issue though.

src/components/documentManager.ts
```diff
@@ -0,0 +1,511 @@
...
+        ) {
+            DocumentManager.showItem(item);
+        });
+    }
+
+    private static getRatingHtml(item: BaseItemDto): string {
+        var html: string = '';
+        if (item.CommunityRating != null) {
+            html +=
+                `<div class="starRating" title="${item.CommunityRating}"></div>` +
```
simonmarklar, line 111: there's no need for string concatenation with template literals - they handle new lines just fine :)  eg:

```
`<div class="starRating" title="${item.CommunityRating}"></div>
<div class="starRatingValue">${item.CommunityRating.toFixed(
  1
)}</div>`
```

hawken93: I've done this previously, but I found the way that prettier split toFixed so obnoxious that I decided to do this to get my own line breaks.

src/components/documentManager.ts
```diff
@@ -0,0 +1,611 @@
...
+        // avoid running it multiple times
+        this.clearBackdropInterval();
+
+        // skip out if it's disabled
+        if (!this.backdropPeriodMs) {
+            this.setWaitingBackdrop(null);
+            return Promise.resolve();
+        }
+
+        this.backdropTimer = <any>(
```
hawken93, line 269: This makes the linter unhappy, but after searching about the setInterval function, I came to the conclusion that it's a mess and I would need help to fix this

hawken93:
Ready for reviews and regression testing!

Afaik this should behave no differently from master

ferferga:
@hawken93 Not strictly related to this PR but I think it fits to be added here:

Right now we "render" the UI for all devices, even Chromecast Audios. Can you add a check here, so if the device is a Chromecast Audio, a ``display: none`` is added to the whole page?

See https://developers.google.com/cast/docs/audio

hawken93:
Hmm I could try. I'm thinking to mess with the detection to make it think my gen2 is an audio for test

src/components/documentManager.ts
```diff
@@ -0,0 +1,611 @@
...
+            '#waiting-container-backdrop'
+        );
+
+        if (element === null) {
+            console.error(
+                'documentManager: Cannot find #waiting-container-backdrop'
+            );
+        } else {
+            (<HTMLElement>element).style.backgroundImage =
+                src != null ? `url(${src})` : '';
```
hawken93, line 214: I don't really like this because the new image loading is visible. I would prefer to preload the new src and then do something like a buffer swap. Someone better at frontend stuff that could help me with this?

src/components/documentManager.ts
```diff
@@ -0,0 +1,611 @@
...
+            if (detailImageUrl != null)
+                detailImageUrl +=
+                    '&PercentPlayed=' +
+                    item.UserData.PlayedPercentage.toString();
+        } else {
+            this.setHasPlayedPercentage(false);
+            this.setPlayedPercentage(0);
+        }
+
+        this.setDetailImage(detailImageUrl);
```
hawken93, line 110: This is the most central code that is responsible for the image on the left hand in the details page. Apparently this image is about half size of what it should be in some cases.

src/components/documentManager.ts
ferferga: @hawken93 I think this is what you want:

https://github.com/jellyfin/jellyfin-vue/pull/408/files#diff-4d53a82f6d48248fa1a22bd2f449b0604526d455c84d8b528ab723182b8f4c76R20

src/components/documentManager.ts
hawken93: Okay I found a css difference between stable and this, turns out there was a css messup.

src/components/documentManager.ts
hawken93: Cool! I'll see what I can do

hawken93: Right. Ty :+1:

hawken93:
@ferferga I've eased up the DOM operations on cc audio devices :)

src/components/documentManager.ts
hawken93: I think we can merge despite this. It's just the only linter warning I couldn't remove.

hawken93:
I have updated the style a bit. I removed the extra h2 text in the waiting container, decreased the h1 to 45px and added a text container bottom right that gets updated with the name of the backdrop item. I've made it so that the text gets updated only when the image has loaded, so that the text switches in sync with the image.

hawken93:
http://puu.sh/H3piL/254a5ec5ae.png

hawken93:
bilde

hawken93:
example

hawken93:
Requesting review and merge. I think this one has worked fine in mine and hamburglar's testing

hamburglar2160:
I've been using this for the past week with no problems. This should get merged when somebody gets a moment to look at it

src/components/documentManager.ts
```diff
@@ -0,0 +1,678 @@
...
+    private static getPrimaryImageUrl(item: BaseItemDto): string | null {
+        if (item.AlbumPrimaryImageTag)
+            return JellyfinApi.createUrl(
+                `Items/${item.AlbumId}/Images/Primary?tag=${item.AlbumPrimaryImageTag}`
+            );
+        else if (item.ImageTags?.Primary)
+            return JellyfinApi.createUrl(
+                `Items/${item.Id}/Images/Primary?tag=${item.ImageTags.Primary}`
+            );
+        else return null;
```
ferferga, line 70: We should probably add a rule for this, but imo all the blocks should have braces. **Don't fix it now, we can do it automatically with eslint in another PR**.

src/components/documentManager.ts
```diff
@@ -0,0 +1,678 @@
...
+                playedIndicator.style.display = 'block';
+                playedIndicator.innerHTML =
+                    '<span class="glyphicon glyphicon-ok"></span>';
+            } else if (item?.UserData?.UnplayedItemCount) {
+                playedIndicator.style.display = 'block';
+                playedIndicator.innerHTML = item.UserData.UnplayedItemCount.toString();
+            } else {
+                playedIndicator.style.display = 'none';
+            }
+        }
```
ferferga, line 133: **This one requires changes**: First of all, I think we should add all the relevant HTML tags to the HTML file itself. The ones that won't appear by default (like this one) can have a ``d-none`` class. I would also create another ``d-block`` css class for ``display: block`` property.

Then, I would create a function like ``toggleVisibility`` which takes an HTMLElement as a parameter and toggles the ``d-none`` class. You're already setting ``display: none`` on initialize() and here, so you can reuse that function. You can also make one for the ``d-block``.

This would lead to better readability.

src/components/documentManager.ts
```diff
@@ -0,0 +1,678 @@
...
+     * Get HTML content used to display the rating of an item
+     *
+     * @param {BaseItemDto} item to look up
+     * @returns {string} html to put in document
+     */
+    private static getRatingHtml(item: BaseItemDto): string {
+        let html = '';
+        if (item.CommunityRating != null) {
+            html +=
+                `<div class="starRating" title="${item.CommunityRating}"></div>` +
```
ferferga, line 189: Same as above

src/components/documentManager.ts
```diff
@@ -0,0 +1,678 @@
...
+                '<div class="starRatingValue">' +
+                item.CommunityRating.toFixed(1) +
+                '</div>';
+        }
+
+        if (item.CriticRating != null) {
+            const verdict = item.CriticRating >= 60 ? 'fresh' : 'rotten';
+            html +=
+                `<div class="${verdict} rottentomatoesicon" title="${verdict}"></div>` +
+                `<div class="criticRating">${item.CriticRating}%</div>`;
```
ferferga, line 199: Same

src/components/documentManager.ts
```diff
@@ -0,0 +1,678 @@
...
+     * @returns {string} html to put in document
+     */
+    private static getRatingHtml(item: BaseItemDto): string {
+        let html = '';
+        if (item.CommunityRating != null) {
+            html +=
+                `<div class="starRating" title="${item.CommunityRating}"></div>` +
+                '<div class="starRatingValue">' +
+                item.CommunityRating.toFixed(1) +
+                '</div>';
```
ferferga, line 192: Same

src/components/documentManager.ts
```diff
@@ -0,0 +1,678 @@
...
+                );
+            } else if (
+                item.ParentBackdropItemId &&
+                item.ParentBackdropImageTags &&
+                item.ParentBackdropImageTags.length
+            ) {
+                // otherwise get first backdrop from parent
+                src = JellyfinApi.createUrl(
+                    `Items/${item.ParentBackdropItemId}/Images/Backdrop/0?tag=${item.ParentBackdropImageTags[0]}`
+                );
```
ferferga, line 254: Can't we also have a function that takes an ID, Image Type and tag and returns us this kind of url without needing it to be scattered all around?

src/components/documentManager.ts
ferferga: This is a number, TypeScript is not wrong about the return type: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setInterval#return_value

src/components/documentManager.ts
```diff
@@ -0,0 +1,678 @@
...
+    public static setPlayerBackdrop(item: BaseItemDto): void {
+        // no backdrop rotation for cc audio
+        if (getActiveDeviceId() === deviceIds.AUDIO) return;
+
+        let backdropUrl: string | null = null;
+
+        if (item.BackdropImageTags && item.BackdropImageTags.length) {
+            backdropUrl = JellyfinApi.createUrl(
+                `Items/${item.Id}/Images/Backdrop/0?tag=${item.BackdropImageTags[0]}`
+            );
```
ferferga, line 381: Same as above

src/components/documentManager.ts
```diff
@@ -0,0 +1,678 @@
...
+        let displayName: string = name;
+
+        if (item.Type == 'TvChannel') {
+            if (item.Number) displayName = `${item.Number} ${name}`;
+        } else if (
+            item.Type == 'Episode' &&
+            item.IndexNumber != null &&
+            item.ParentIndexNumber != null
+        ) {
+            let episode = `S${item.ParentIndexNumber}, E${item.IndexNumber}`;
```
ferferga, line 456: Unrelated, but we should remember about this so we translate it once #111 is merged

src/index.html
```diff
@@ -18,9 +18,9 @@
     <div id="waiting-container">
         <div class="waitingContent">
             <img class="logo" src="img/logo.svg" />
-            <h1>Ready to cast</h1>
-            <h2>Select your media in Jellyfin and play it here</h2>
+            Ready to cast
```
ferferga, line 21: Why remove ``h1``?

ferferga (changes requested):

src/components/documentManager.ts
hawken93: It seems like this one just takes backdrop number 0, ideally we would pick a random one, I'd like to express that too, somehow

src/components/documentManager.ts
hawken93: Typescript thinks this is not a number

src/index.html
hawken93: It forces newline, alternately I could have tried the image within a h1 tag, but ultimately the css styling provides everthing we need

src/components/documentManager.ts
hawken93: we have a .hide class, not sure if it's used

src/components/documentManager.ts
hawken93:
```
ERROR in /home/hawken/git/mirror/jellyfin-chromecast/src/components/documentManager.ts
./components/documentManager.ts
[tsl] ERROR in /home/hawken/git/mirror/jellyfin-chromecast/src/components/documentManager.ts(352,9)
      TS2322: Type 'Timeout' is not assignable to type 'number'.
 @ ./components/maincontroller.ts 8:0-52 91:8-45
 @ ./app.ts 2:0-37
```

hawken93: https://stackoverflow.com/questions/45802988/typescript-use-correct-version-of-settimeout-node-vs-window

hawken93: This issue is beyond me, I really don't want to do some other hack instead of this hack. If we are going for a hack I'll add a comment here to tell the linter to leave it alone.

src/components/documentManager.ts
hawken93: Fixed it by moving this stuff straight into html

hawken93:
Okay @hamburglar2160 request new test, I've only deployed this PR right now.

@ferferga  @simonmarklar new review?

hawken93:
I've modified waiting and details pages to use relative scaling, should make 4k display look the same as 720p display and I think it's for the better, @hamburglar2160 opinion?

src/components/deviceprofileBuilder.ts
```diff
@@ -41,11 +41,11 @@ let profileOptions: ProfileOptions;
 let currentDeviceId: number;
 
 /**
- * @param Property What property the condition should test.
- * @param Condition The condition to test the values for.
- * @param Value The value to compare against.
- * @param [IsRequired=false]
```
ferferga, line 47: Why the default value is removed?

src/components/documentManager.ts
```diff
@@ -0,0 +1,778 @@
...
+        return this.preloadImage(src);
+    }
+
+    /**
+     * This fucntion takes an item and shows details about it
+     * on the details page. This happens when no media is playing,
+     * and the connected client is browsing the library.
+     *
+     * @param {BaseItemDto} item to show information about
+     * @returns {Promise<void>} for the page to load
```
ferferga, line 133: JSDoc syntax is like this: @param {type} argumentInFunction - Description

@param {type} [argumentInFunction=value] - Description

for arguments that have default values.

Non-blocking though, I can fix that in the linting PR.

src/components/deviceprofileBuilder.ts
hawken93: I didn't understand how this would fit in to the jsdoc syntax, from the other functions I didn't see any practice of documenting default values with this syntax. Feel free to suggest a change

src/components/documentManager.ts
hawken93: As far as I can see the hyphen is optional and just a readability thing, but from this I can fix the other one

src/components/deviceprofileBuilder.ts
hawken93: Okay, fixed this :)

ferferga (approved):
Assumming this was fully tested after the new changes, this LGTM

PR #95, opened 2020-12-13
Add tests to credentialManager

YouKnowBlom:
Add constructor to `credentialManager` with optional parameter to initialize the store with.
Add tests for methods in `credentialManager`.
Remove `--passWithNoTests` from test command to disallow passing when no tests are found.

### DIFF
--- a/src/__tests__/credentialManager.test.ts
+++ b/src/__tests__/credentialManager.test.ts
@@ -1,0 +1,88 @@
+import { Configuration } from "../api/generated/configuration";
+
+import {
+    credentialManager,
+    CredentialStore
+} from "../api/credentialManager";
+
+const serverId1 = "f4486b851af24255b3305fe614b81f01";
+const serverConfig1: Configuration = {
+    apiKey: "b49268e51af24255b3305fe614b81f01"
+};
+
+const serverId2 = "af72hb851af24255b3305fe614b81f01";
+const serverConfig2: Configuration = {
+    apiKey: "d4286b8119f24a55b3305fe614b81f01"
+};
+
+describe("Getting servers from credentialManager", () => {
+    let credentialMgr: credentialManager;
+
+    beforeEach(() => {
+        credentialMgr = new credentialManager({
+            [serverId1]: serverConfig1
+        });
+    });
+
+    test("Get existing server from store", () => {
+        expect(credentialMgr.get(serverId1)).toEqual(serverConfig1);
+    });
+
+    test("Get non-existant server from store", () => {
+        expect(credentialMgr.get(serverId2)).toBeUndefined();
+    });
+});
+
+describe("Updating servers in credentialManager", () => {
+    let credentialMgr: credentialManager;
+
+    beforeEach(() => {
+        credentialMgr = new credentialManager({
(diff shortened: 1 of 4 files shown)


src/__tests__/credentialManager.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    let credentialMgr: credentialManager;
+
+    beforeEach(() => {
+        credentialMgr = new credentialManager({
+            [serverId1]: serverConfig1
+        });
+    });
+
+    test("Add server id to store", () => {
+        expect(credentialMgr.add(serverId2, serverConfig2)).toBeTruthy() ;
```
dkanada, line 64:
```suggestion
    test("Add server ID to store", () => {
        expect(credentialMgr.add(serverId2, serverConfig2)).toBeTruthy();
```

dkanada: seems like this should have been caught by linting

src/__tests__/credentialManager.test.ts
```diff
@@ -0,0 +1,88 @@
...
+const serverConfig1: Configuration = {
+    apiKey: "b49268e51af24255b3305fe614b81f01"
+};
+
+const serverId2 = "af72hb851af24255b3305fe614b81f01";
+const serverConfig2: Configuration = {
+    apiKey: "d4286b8119f24a55b3305fe614b81f01"
+};
+
+describe("Getting servers from credentialManager", () => {
```
dkanada, line 18: I prefer all lowercase for tests, but since I'm not working on this project I won't actually request that change haha.

dkanada:
@YouKnowBlom this one also seems good to merge if you rebase against master and optionally address the comments.

src/__tests__/credentialManager.test.ts
dkanada: I think you might have missed the stray space character here.

dkanada:
Is this still failing after you rebased against master? I'm less keen on merging it if the new tests will break the CI workflows.

src/__tests__/credentialManager.test.ts
YouKnowBlom: Oh whoops! Will fix that soon

YouKnowBlom:
Yeah it is broken right now. The fix is setting some values in the tsconfig but I haven't gotten around to it yet, I'll let you know when it's done

YouKnowBlom:
Now it should be working! I fixed the linting issues in the related files but there are many unrelated warnings which will have to go into a seperate PR.

dkanada:
Why is the linting failing now out of nowhere? There aren't really any changes to enable new checks here. If nothing else, I would rather disable the checks that are throwing errors until they're actually fixed in the codebase to avoid collecting new issues.

YouKnowBlom:
I'm confused as well. I'll take a closer look tonight to see if I can figure it out. If not I'll disable them for now

YouKnowBlom:
Apparently the prettier was mad that the formatting changed in tsconfig and was returning 1, fixed now! @dkanada

dkanada (approved):

PR #97, opened 2020-12-14
Add Webpack watch and dev server scripts

simonmarklar:
- added `yarn serve`/ `yarn start` command using webpack dev server
- added `yarn watch` for those using their own web server
- added fixpack to keep `package.json` uniform

Please note that while i've tested that it builds and launches/displays in a browser im still having issues with jellyfin detecting my chromecast so cant fully test :'(


YouKnowBlom:
Is the webpack config large enough to warrant 4 files

simonmarklar:
> Is the webpack config large enough to warrant 4 files

each file has a different purpose and keeps the options for each purpose in one place. I find it easier to grok this way but if you prefer it to be in one file with command line args i can do that :)

simonmarklar:
@YouKnowBlom i've merged the configs down to one. Webpack 5 has a bunch of new stuff which makes this really neat :) also figured i'd change the config over to typescript seeing as the rest of the codebase is headed that way

YouKnowBlom:
Awesome, I hope I didn't discourage you :p. Looks very neat now! Will take an extra look when I get home :+1: 

It's a busy time of the year so excuse me for taking a while to reply.

simonmarklar:
all good mate - I've not looked at webpack 5 before this ticket so I just went with what I am familiar with, reading the docs really helped on one of my insomnia nights ;)

all good on reply speed - this isnt our jobs ;)

ferferga:
@simonmarklar Can you rebase this please? Last PR that converts everything to TS ^^.

src/components/jellyfinActions.ts
```diff
@@ -180,7 +180,7 @@ export function pingTranscoder(
     if (now - lastTranscoderPing < 5000) {
         console.debug('Skipping ping due to recent progress check-in');
         return new Promise(function (resolve) {
-            resolve();
+            resolve(undefined);
         });
```
hawken93, line 184: hmm should this be `return Promise.resolve()`?

tsconfig.json
```diff
@@ -1,10 +1,12 @@
 {
     "compilerOptions": {
         "target": "ES2018",
-        "module": "ESNext",
+        "module": "CommonJS",
```
ferferga, line 4: Why?

simonmarklar: ah crap, thats not supposed to be there. got an error about importing `webpack.DefinePlugin` in the config file and was testing things trying to get rid of it... good catch :)

src/components/jellyfinActions.ts
simonmarklar: There's a lot of places in this codebase that should probably be doing that... Putting in that undefined cleared out one error and i realized that I probably shouldn't fix the errors in this PR ... honestly didn't think I included this file in the PR but i must of

simonmarklar:
> @simonmarklar Can you rebase this please? Last PR that converts everything to TS ^^.

OK that should be all done.

I'm not 100% sold on the webpack config being typescript, there was a few hoops to jump through. What's peoples thoughts? Its nice having the typings for config but it comes at a cost of several dev dependencies

tsconfig-webpack.json
```diff
@@ -0,0 +1,8 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
```
hawken93, line 3: Should this be ESNext? I have no idea about this stuff

.eslintrc.js
```diff
@@ -7,14 +7,14 @@ module.exports = {
     },
```
hawken93, line 7: For the eslintrc environment, I had to add browser: true. Maybe remove node: true here, and add browser: true because of the override in the webpack config.

hawken93 (changes requested):
+ my comment in jellyfinActions.ts, but I think I'm going to do a refactoring to async syntax later.

Would like changes for the two other comments if you agree with them.

tsconfig-webpack.json
simonmarklar: no. This file is explicitly for `webpack.config.ts`   - it will not work at all if its not `commonjs`

https://webpack.js.org/configuration/configuration-languages/#typescript

.eslintrc.js
simonmarklar: that seems to be done further down in the `overrides` section. 

I do need to move the `prettier/recommended` plugin to the end of the array though

simonmarklar:
@hawken93 

> plus my comment in jellyfinActions.ts, but I think I'm going to do a refactoring to async syntax later.

i dont want to refactor the code in `jellfinActions.ts` in this PR - I only put the undefined in there to appease eslint and make it so I could run builds and verify that things still worked. 

There's a ton of refactoring that needs to be done to bring the code in line but IMHO we're better off doing focused PRs - plus i see so much in progress work and I know you're smashing things out so im willing to bet its already been done by someone :)

> Would like changes for the two other comments if you agree with them.

I dont think they are necessary - see my comments on them for more info :)

ferferga (approved):
Sorry for taking so long in getting this merged. Thank you very much for your contribution!

hawken93 (approved):
Looks good to me, I have gotten my questions answered. Would probably have preferred the promise resolve() to not be altered at all to make sure the linter still notes it as a problem. But that's very minor, especially with additional checks coming in that will cause us to rewrite away from promises 99% of cases

Issue #80, opened 2020-11-28
Typescript migration


YouKnowBlom:
Opening this issue to track progress on the ongoing typescript migration:

- [x] `src/app.js`
- [x] `src/helpers.js`
- [x] `src/components/castDevices.js`
- [x] `src/components/codecSupportHelper.js`
- [x] `src/components/commandHandler.js`
- [x] `src/components/deviceprofileBuilder.js`
- [x] `src/components/fetchhelper.js`
- [x] `src/components/jellyfinActions.js`
- [x] `src/components/maincontroller.js`
- [x] `src/components/playbackManager.js`

neopc10:
\o/

ferferga:
Finished on #99, needs to add typings and JSDoc. Partial work done on #106

YouKnowBlom:
Closing this as all boxes have been ticked 🚀

PR #99, opened 2020-12-21
Conversions to typescript and misc stuff

hawken93:
* move linting to root dir, lint js and ts with lint:code, and update workflow to use it ignore api/generated, not the whole api/ folder
* commandHandler doesn't need to do that cast API import
* Playbackmanager to ts
* order package.json alphabetically and make `yarn serve` a thing
* Stop requests happening when the server is unknown, also move url creation into the ajax function.
* maincontroller to ts
* fetchhelper to ts
* helpers to ts
* Try fixing system volume call, it called a function that seems to not exist now. Couldn't find it in the docs. Not sure if it was broken before. Fixed up some callbacks to populate window.volume to get this info
* application ready message moved down to the call that activates CAF (Still doesn't wait for it to be ready before doing other stuff though)

Please test throughly! A lot changed and regressions are possible..


src/components/commandHandler.ts
```diff
@@ -27,114 +26,114 @@ interface SupportedCommands {
     [command: string]: (data: DataMessage) => any;
 }
 
-export class commandHandler {
```
hawken93, line 30: This class didn't need instantiation, we were just keeping one instance in window.commandHandler before.

src/components/fetchhelper.ts
```diff
@@ -30,36 +30,36 @@ export function getFetchPromise(request) {
         : fetch(url, fetchRequest);
 }
 
-export function fetchWithTimeout(url, options, timeoutMs) {
-    return (
-        console.log(
-            'fetchWithTimeout: timeoutMs: ' + timeoutMs + ', url: ' + url
-        ),
```
hawken93, line 37: refactored this stupid statement here, leftovers from when it was minimized

YouKnowBlom:
> Try fixing system volume call, it called a function that seems to not exist now. Couldn't find it in the docs. Not sure if it was broken before. Fixed up some callbacks to populate window.volume to get this info

We shouldn't have to change this on the sender anymore I believe

hawken93:
Lol it's failing the tests because there are no js files left :D

src/helpers.ts
```diff
@@ -10,13 +14,13 @@ export function getCurrentPositionTicks($scope) {
...
 
-export function getReportingParams($scope) {
+export function getReportingParams($scope: GlobalScope): any {
     /* Math.round() calls:
      * on 10.7, any floating point will give an API error,
      * so it's actually really important to make sure that
      * those fields are always rounded.
      */
-    var volumeInfo = window.castReceiverContext.getSystemVolume();
+    var volumeInfo = window.volume;
```
hawken93, line 23: In the code below here we are reporting back the the server the volume of the cast device, so if I understand correctly, we don't need this?

hawken93:
#80

hawken93:
have made `yarn lint:ts` possible and fixed lint

src/api/credentialManager.ts
```diff
@@ -1,10 +1,10 @@
-import {Configuration} from "./generated/configuration"
+import { Configuration } from './generated/configuration';
```
Artiume, line 1:
```suggestion
import { Configuration } from 'generated/configuration';
```

would this work?

hawken93: I don't think so, any path without the leading dot is treated like an external module (node_modules)

hawken93: I found some documentation if you are interested at https://www.typescriptlang.org/docs/handbook/module-resolution.html

src/helpers.ts
hawken93: ...we should probably let jellyfin in on what the chromecast volume is

hawken93:
Tested this watching a movie and it worked allright :)

.eslintrc.js
```diff
@@ -48,7 +48,9 @@ module.exports = {
             rules: {
                 // Disable these until we have converted the project to TS
                 '@typescript-eslint/explicit-module-boundary-types': 'off',
-                '@typescript-eslint/no-explicit-any': 'off'
+                '@typescript-eslint/no-explicit-any': 'off',
```
ferferga, line 51:
```suggestion
```

We must not allow any rules that can potentially get bad code in. Much better to ignore it for specific lines imo.
See comments for suggestions in how to deal with this.

src/components/fetchhelper.ts
```diff
@@ -1,12 +1,12 @@
-export function getFetchPromise(request) {
-    var headers = request.headers || {};
+export function getFetchPromise(request: any): Promise<any> {
```
ferferga, line 1:
```suggestion
export function getFetchPromise(request: never): Promise<never> {
```

If it's 100% impossible to determine exactly what kind of objects we will get here, use ``never`` or ``unknown``. As more things we have without strictly typing, more bugs can slip.

src/components/fetchhelper.ts
```diff
@@ -20,49 +20,49 @@ export function getFetchPromise(request) {
...
-                    reject();
-                }
-            );
-        })
-    );
+export function fetchWithTimeout(
+    url: string,
+    options: any,
+    timeoutMs: number
+): Promise<any> {
```
ferferga, line 37:
```suggestion
export function fetchWithTimeout(
    url: string,
    options: Record<string, string>,
    timeoutMs: number
): Promise<void> {
```

Same as above. I'm on phone and can't say from memory, but iirc, this should be a dictionary with keys and values of type string, so my suggestion applies. If not, send some examples of possible dicts and I can help you get the correct type.

Also, use ``void`` as return value on functions that doesn't return anything.

src/components/fetchhelper.ts
```diff
@@ -73,13 +73,13 @@ export function paramsToString(params) {
     return values.join('&');
 }
 
-export function ajax(request) {
+export function ajax(request: any): Promise<any> {
```
ferferga, line 76: Same as above

src/components/fetchhelper.ts
```diff
@@ -73,13 +73,13 @@ export function paramsToString(params) {
...
     if (!request) throw new Error('Request cannot be null');
     request.headers = request.headers || {};
     console.log('requesting url: ' + request.url);
 
     return getFetchPromise(request).then(
-        function (response) {
+        function (response: any) {
```
ferferga, line 82: Same as above. This could be an arrow function as well.

src/components/jellyfinActions.ts
```diff
@@ -67,17 +67,16 @@ export function reportPlaybackStart(
...
-
-    broadcastToMessageBus({
+    broadcastToMessageBus(
         //TODO: convert these to use a defined type in the type field
-        type: 'playbackstart',
-        data: getSenderReportingData($scope, reportingParams)
-    });
+        'playbackstart',
+        undefined,
+        getSenderReportingData($scope, reportingParams)
```
ferferga, line 74: Why this is no longer an options object. The undefined in the middle seems ugly.

src/components/jellyfinActions.ts
```diff
@@ -87,24 +86,23 @@ export function reportPlaybackStart(
...
 ): Promise<any> {
-    broadcastToMessageBus({
-        type: broadcastEventName || 'playbackprogress',
-        data: getSenderReportingData($scope, reportingParams)
-    });
+    broadcastToMessageBus(
+        broadcastEventName,
+        undefined,
+        getSenderReportingData($scope, reportingParams)
+    );
```
ferferga, line 96: Same as above

src/components/jellyfinActions.ts
```diff
@@ -117,14 +115,13 @@ export function reportPlaybackStopped(
...
-
-    broadcastToMessageBus({
-        type: 'playbackstop',
-        data: getSenderReportingData($scope, reportingParams)
-    });
+    broadcastToMessageBus(
+        'playbackstop',
+        undefined,
+        getSenderReportingData($scope, reportingParams)
+    );
```
ferferga, line 122: Same as above

src/components/jellyfinActions.ts
```diff
@@ -269,24 +264,14 @@ function showItem(item: BaseItemDto): void {
...
 
 export function displayItem(itemId: string): void {
-    const url = JellyfinApi.createUserUrl('Items/' + itemId);
-
-    JellyfinApi.authAjax(url, {
+    JellyfinApi.authAjaxUser('Items/' + itemId, {
         dataType: 'json',
         type: 'GET'
-    }).then(function (item) {
+    }).then(function (item: BaseItemDto) {
```
ferferga, line 270: This could be an arrow function as well

src/components/jellyfinActions.ts
```diff
@@ -379,8 +362,8 @@ export function getLiveStream(
     deviceProfile: DeviceProfile,
     startPosition: number,
     mediaSource: MediaSourceInfo,
-    audioStreamIndex: number,
-    subtitleStreamIndex: number
+    audioStreamIndex: number | null,
+    subtitleStreamIndex: number | null
 ): Promise<any> {
```
ferferga, line 367:
```suggestion
): Promise<void> {
```

src/components/jellyfinApi.ts
```diff
@@ -75,9 +71,39 @@ export abstract class JellyfinApi {
     }
 
     // Authenticated ajax
-    public static authAjax(url: string, args: any): Promise<any> {
+    public static authAjax(path: string, args: any): Promise<any> {
```
ferferga, line 74:
```suggestion
    public static authAjax(path: string, args: any): Promise<void> {
```

Same as above with all the ``any``'s

src/components/maincontroller.ts
```diff
@@ -194,33 +203,30 @@ window.mediaManager.addEventListener(
...
+        return JellyfinApi.authAjax('Sessions/Capabilities/Full', {
             type: 'POST',
             data: JSON.stringify(capabilities),
             contentType: 'application/json'
         });
     });
 }
 
-export function processMessage(data) {
+export function processMessage(data: any): void {
```
ferferga, line 229: Same as above

src/components/maincontroller.ts
```diff
@@ -296,28 +291,34 @@ export function processMessage(data) {
     }
 }
 
-export function reportEvent(name, reportToServer) {
-    reportPlaybackProgress(
+export function reportEvent(
+    name: string,
+    reportToServer: boolean
+): Promise<any> {
```
ferferga, line 297: Same as above

src/components/maincontroller.ts
```diff
@@ -372,40 +373,45 @@ export function setSubtitleStreamIndex($scope, index) {
     }
 }
 
-export function setAudioStreamIndex($scope, index) {
-    var positionTicks = getCurrentPositionTicks($scope);
-    changeStream(positionTicks, {
+export function setAudioStreamIndex(
+    $scope: GlobalScope,
+    index: number
+): Promise<void> {
```
ferferga, line 379: Same as above

src/components/maincontroller.ts
```diff
@@ -584,24 +596,20 @@ export function getMaxBitrate() {
...
-        type: 'playbackerror',
-        message: errorCode
-    });
+export function showPlaybackInfoErrorMessage(error: string): void {
+    broadcastToMessageBus('playbackerror', error);
 }
 
-export function getOptimalMediaSource(versions) {
-    var optimalVersion = versions.filter(function (v) {
+export function getOptimalMediaSource(versions: Array<any>): any {
```
ferferga, line 611: Same as above

src/components/maincontroller.ts
```diff
@@ -463,12 +469,12 @@ export function changeStream(ticks, params) {
...
     'urn:x-cast:com.connectsdk',
-    function (evt) {
-        var data = evt.data;
+    function (evt: any) {
+        let data: any = evt.data;
 
         // Apparently chromium likes to pass it as json, not as object.
         // chrome on android works fine
-        if (typeof data == 'string' || data instanceof String) {
+        if (typeof data == 'string') {
```
ferferga, line 477:
```suggestion
        if (typeof data === 'string') {
```

If it doesn't exist, we should make a rule in this repo for enforcing this.

src/components/maincontroller.ts
```diff
@@ -566,7 +578,7 @@ export function getMaxBitrate() {
 
         console.log('detecting bitrate');
 
-        detectBitrate($scope).then(
+        detectBitrate().then(
             function (bitrate) {
```
ferferga, line 582: Another thing that could be an arrow function

src/components/maincontroller.ts
```diff
@@ -584,24 +596,20 @@ export function getMaxBitrate() {
     });
 }
 
-export function validatePlaybackInfoResult(result) {
+export function validatePlaybackInfoResult(result: any): boolean {
```
ferferga, line 599: Same as above

src/components/maincontroller.ts
```diff
@@ -621,7 +629,8 @@ export function getOptimalMediaSource(versions) {
     );
 }
 
-export function supportsDirectPlay(mediaSource) {
+// TODO wtf and why do we need this, direct storage access but how
+export function supportsDirectPlay(mediaSource: any): boolean {
```
ferferga, line 633: Same as above

src/components/maincontroller.ts
```diff
@@ -701,8 +716,13 @@ export function setTextTrack(index) {
     }
 }
 
-export function createMediaInformation(playSessionId, item, streamInfo) {
-    var mediaInfo = new cast.framework.messages.MediaInformation();
+// TODO no any types
+export function createMediaInformation(
+    playSessionId: string,
+    item: BaseItemDto,
+    streamInfo: any
```
ferferga, line 723: Same as above

src/components/maincontroller.ts
```diff
@@ -769,7 +789,7 @@ if (!PRODUCTION) {
 
     window.mediaManager.addEventListener(
         cast.framework.events.category.CORE,
-        (event) => {
+        (event: any) => {
```
ferferga, line 792:
```suggestion
        (event: Event) => {
```

src/components/playbackManager.ts
```diff
@@ -55,71 +62,77 @@ export class playbackManager {
         );
     }
 
-    async playFromOptions(options) {
+    async playFromOptions(options: any): Promise<boolean> {
```
ferferga, line 65: Same as above

src/components/playbackManager.ts
```diff
@@ -55,71 +62,77 @@ export class playbackManager {
...
-        let intros = await getIntros(firstItem);
-        options.items = intros.Items.concat(options.items);
-        this.playFromOptionsInternal(options);
+        const intros = await getIntros(firstItem);
+        options.items = intros.Items?.concat(options.items);
+        return this.playFromOptionsInternal(options);
     }
 
-    playFromOptionsInternal(options) {
+    playFromOptionsInternal(options: any): boolean {
```
ferferga, line 77: Same as above

src/components/playbackManager.ts
```diff
@@ -55,71 +62,77 @@ export class playbackManager {
...
+        window.currentPlaylistIndex = -1;
         window.playlist = this.activePlaylist;
 
-        this.playNextItem(options, stopPlayer);
+        return this.playNextItem(options, stopPlayer);
     }
 
-    playNextItem(options, stopPlayer) {
-        var nextItemInfo = getNextPlaybackItemInfo();
+    playNextItem(options: any = {}, stopPlayer = false): boolean {
```
ferferga, line 88: Same as above

src/components/playbackManager.ts
```diff
@@ -55,71 +62,77 @@ export class playbackManager {
...
-            this.playItem(item, options || {}, stopPlayer);
+            this.playItem(item, options, stopPlayer);
             return true;
         }
 
         return false;
     }
 
-    playPreviousItem(options) {
+    playPreviousItem(options: any = {}): boolean {
```
ferferga, line 103: Same as above

src/components/playbackManager.ts
```diff
@@ -55,71 +62,77 @@ export class playbackManager {
...
+            this.playItem(item, options, true);
             return true;
         }
         return false;
     }
 
-    async playItem(item, options, stopPlayer) {
+    async playItem(
+        item: BaseItemDto,
+        options: any,
```
ferferga, line 117: Same as above

ferferga (changes requested):
Lots of ``any``'s around

.eslintrc.js
hawken93: I don't think we can remove this until it's feasible to get a clean lint. There are lots of types that need to be set up and I don't want this PR to gets much bigger because there's already too much that can break...

src/components/fetchhelper.ts
hawken93: It returns stuff.
I have started modelling the request type because it's a structure with a limited scope. I'm confident enough to refactor that now

hawken93: And I think options.headers is its own dictionary.

src/components/fetchhelper.ts
hawken93: These return a fetch response. I see tutorials using templating for the response type, for the arguments I suspect there is something ready to use.

src/components/fetchhelper.ts
hawken93: Here the request is a custom type that needs its own definition. The response will be interpreted to various objects througout the code, so it should really be templated somehow.

src/components/jellyfinApi.ts
hawken93: We use the result, sorry

src/components/maincontroller.ts
hawken93: I'm not sure but I think I put this one to void now

src/components/maincontroller.ts
hawken93: This one will need a new interface I believe

src/components/maincontroller.ts
hawken93: This function is going away soon

src/components/maincontroller.ts
hawken93: The whole streamInfo construct is quite random to me, I think they just put whatever in there so it's definitely a new type

src/components/playbackManager.ts
hawken93: This deserves a new interface as well

src/components/playbackManager.ts
```diff
@@ -167,19 +176,29 @@ export class playbackManager {
         );
     }
 
-    playMediaSource(playSessionId, item, mediaSource, options) {
+    // TODO eradicate any
+    playMediaSource(
+        playSessionId: string,
+        item: BaseItemDto,
+        mediaSource: any,
```
hawken93, line 183: I think this one can be set to the MediaSourceInfo type if I remember correctly

src/components/jellyfinActions.ts
hawken93: Undid that after I understood how interfaces work

src/components/jellyfinActions.ts
hawken93: These are in use

src/components/maincontroller.ts
hawken93: I think there's a type in cast.framework that is more specific here

src/components/fetchhelper.ts
hawken93: I did find better types for this function

src/components/fetchhelper.ts
hawken93: Fixed half of it

src/components/maincontroller.ts
hawken93: Depends on a proper type for streamInfo

src/components/maincontroller.ts
hawken93: yey :)

src/components/maincontroller.ts
hawken93: seek PR shows how this will be removed

hawken93:
I have decided that I will not be refactoring more `any` types if they require making new interfaces in this PR.
I have started looking at queue manager stuff, and this will anyway require me to take a serious new look at all of the complex structures that are getting passed around.

ferferga (approved):
As commented many times in chat, we can fix linting and add more typings in the future, otherwise this PR will be insane (it's already insane, but we don't want to make it more insane) :).

I checked your newly added typings and I would say the status of this PR is much better than what it was when I requested changes, so as long as it works and doesn't break the client completely, I would say it's good to go.

Ping me once all the client testing is done.

src/components/playbackManager.ts
ferferga: @hawken93 Cool! Remember to change this before merging please.

src/components/maincontroller.ts
hawken93: `(event: cast.framework.events.MediaElementEvent)`

hawken93: Hm actually you're right, the closest match seems to be Event.

hawken93: Hm. cast.framework.events.Event

hawken93:
Okay, I believe at this point @ferferga is satisfied with the changes after his review, despite not dealing with everything. I will now stop any more updates to this PR to allow for testing :)

hamburglar2160:
Regression: Admin Dashboard->Active Devices does not report playback
Server/web: 10.6.2

hamburglar2160:
Regression: Can't seem to skip around music tracks from the Cast framework side and apparently web, but need to investigat efurther

hawken93:
I'm reverting my attempt to fix the playback reporting because this code is messy as heck. Trying to change it really opens a can of worms.

hawken93:
It still behaves differently from nightly. Start playing an item and this PR doesn't show the progress bar for some reason

hawken93:
Okay.. I think I got that bug fixes, it was in helpers.ts: getSenderReportingData. After dropping the extend() call it was not a shallow copy anymore. The null & delete operations deleted that information for everyone I think.

I figured it would be easier to not take a copy and live with sending some extra fields.

hamburglar2160:
I've tested this on a whole bunch of devices and am not finding any regressions. LGTM

hawken93:
Asking for anyone that would like to look over the code once more or test it :)

PR #100, opened 2020-12-21
WIP: enable mkv and enable passthrough codecs for mkv

hawken93:
This enables mkv. This is not ideal because chromecast does not officially support mkv and there are edge cases where chromecast will choke on some of these mkv files.

However, this is how it works on chromecast stable, it enables AC3 and EAC3 to work and it doesn't suffer from the performance issues that I have seen while trying to play HLS.

This should be seen as buying time to get HLS/fmp4 or an alternative that is actually supported working.

EDIT: this branch works with eac3+h264 on gen2+gen3 at least.
Please test this a lot :)

### DIFF
--- a/src/components/deviceprofileBuilder.ts
+++ b/src/components/deviceprofileBuilder.ts
@@ -72,14 +72,7 @@ function getContainerProfiles(): Array<ContainerProfile> {
  * @returns Response profiles.
  */
 function getResponseProfiles(): Array<ResponseProfile> {
-    // This seems related to DLNA, it might not be needed?
-    return [
-        {
-            Container: 'm4v',
-            MimeType: 'video/mp4',
-            Type: DlnaProfileType.Video
-        }
-    ];
+    return [];
 }
 
 /**
--- a/src/components/deviceprofileBuilder.ts
+++ b/src/components/deviceprofileBuilder.ts
@@ -16,7 +16,6 @@ import { ProfileConditionValue } from '../api/generated/models/profile-condition
 import { deviceIds, getActiveDeviceId } from './castDevices';
 
 import {
-    hasSurroundSupport,
     hasTextTrackSupport,
     hasVP8Support,
     hasVP9Support,
@@ -28,6 +27,7 @@ import {
     getSupportedMP4AudioCodecs,
     getSupportedHLSVideoCodecs,
     getSupportedHLSAudioCodecs,
+    getSupportedSurroundCodecs,
     getSupportedWebMAudioCodecs,
     getSupportedAudioCodecs
 } from './codecSupportHelper';
@@ -89,8 +89,10 @@ function getDirectPlayProfiles(): Array<DirectPlayProfile> {
     const DirectPlayProfiles: Array<DirectPlayProfile> = [];
 
     if (currentDeviceId !== deviceIds.AUDIO) {
+        // For devices with video
         const mp4VideoCodecs = getSupportedMP4VideoCodecs();
         const mp4AudioCodecs = getSupportedMP4AudioCodecs();
+        const surroundAudioCodecs = getSupportedSurroundCodecs();
         const vpxVideoCodecs = getSupportedVPXVideoCodecs();
         const webmAudioCodecs = getSupportedWebMAudioCodecs();
 
@@ -103,6 +105,15 @@ function getDirectPlayProfiles(): Array<DirectPlayProfile> {
             });
         }
 
+        // mkv profile: surround and normal codecs
+        DirectPlayProfiles.push({
+            AudioCodec: mp4AudioCodecs.concat(surroundAudioCodecs).join(','),
+            Container: 'mkv',
+            Type: DlnaProfileType.Video,
+            VideoCodec: mp4VideoCodecs.join(',')
+        });
+
+        // HLS+MPEGTS profile
         DirectPlayProfiles.push({
             AudioCodec: mp4AudioCodecs.join(','),
             Container: 'mp4,m4v',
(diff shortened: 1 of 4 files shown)


src/components/deviceprofileBuilder.ts
```diff
@@ -267,32 +278,48 @@ function getTranscodingProfiles(): Array<TranscodingProfile> {
...
+
+    // Passthrough for audio only
+    if (surroundCodecs.length) {
+        // surround audio, try mkv here
+        TranscodingProfiles.push({
+            Container: 'mkv',
+            Type: DlnaProfileType.Audio,
+            AudioCodec: surroundCodecs.join(','),
+            Context: EncodingContext.Streaming,
+            Protocol: 'http'
```
hawken93, line 322: I have no idea if this will work, but I'm trying to tell jellyfin to send me an ac3 audio file muxed in mkv

src/components/deviceprofileBuilder.ts
```diff
@@ -302,6 +329,31 @@ function getTranscodingProfiles(): Array<TranscodingProfile> {
...
+    if (hlsVideoCodecs.length && surroundCodecs.length) {
+        // Direct streaming for passthrough codecs
+        TranscodingProfiles.push({
+            Container: 'mkv',
+            Type: DlnaProfileType.Video,
+            AudioCodec: surroundCodecs.join(','),
+            VideoCodec: hlsVideoCodecs.join(','),
+            Context: EncodingContext.Streaming,
+            Protocol: 'http'
+            //MaxAudioChannels: 6 For passthrough it's not really relevant, trying without here.
```
hawken93, line 342: No idea if this will work either, Seems silly to limit audio channels when the chromecast doesn't even care what the channels are

hawken93: Seems like it didn't hurt to leave it out

src/components/deviceprofileBuilder.ts
```diff
@@ -104,6 +106,15 @@ function getDirectPlayProfiles(): Array<DirectPlayProfile> {
             });
         }
 
+        // mkv profile: surround and normal codecs
```
hawken93, line 109: I haven't commented clearly enough that chromecast does not really support this and it may fail on some files.

src/components/deviceprofileBuilder.ts
```diff
@@ -267,32 +278,48 @@ function getTranscodingProfiles(): Array<TranscodingProfile> {
...
             AudioCodec: audioFormat,
             Context: EncodingContext.Streaming,
             Protocol: 'http',
-            MaxAudioChannels: audioChannels.toString()
+            // only 2ch supported for audio
+            MaxAudioChannels: '2'
+        });
+    }
+
+    // Passthrough for audio only
```
hawken93, line 314: I doubt this is going to meet the edge case of an mkv audio only file that is incompatible. Sounds very unlikely in this instance

src/components/deviceprofileBuilder.ts
```diff
@@ -302,6 +329,31 @@ function getTranscodingProfiles(): Array<TranscodingProfile> {
     }
 
     const hlsVideoCodecs = getSupportedHLSVideoCodecs();
+
+    if (hlsVideoCodecs.length && surroundCodecs.length) {
+        // Direct streaming for passthrough codecs
```
hawken93, line 334: I haven't commented clearly enough that chromecast does not really support this and it may fail on some files.

src/components/deviceprofileBuilder.ts
```diff
@@ -302,6 +329,31 @@ function getTranscodingProfiles(): Array<TranscodingProfile> {
...
+            Container: 'mkv',
+            Type: DlnaProfileType.Video,
+            AudioCodec: surroundCodecs.join(','),
+            VideoCodec: hlsVideoCodecs.join(','),
+            Context: EncodingContext.Streaming,
+            Protocol: 'http'
+            //MaxAudioChannels: 6 For passthrough it's not really relevant, trying without here.
+        });
+
+        // Direct streaming for normal codecs
```
hawken93, line 345: same as over

hawken93:
IMO the only sustainable way to do MKV is to do it with directplay.

Remuxing into mkv works fine, but seeking will never work properly with it, neither will the progress bar. I kinda want to keep it for experimentation but it's not good. Speaks to the need for figuring out HLS properly.

hawken93:
Please don't merge yet, couldn't mark it draft

hawken93:
I think it can be enabled for direct play at least, this gives a good experience. For transcoding/remux it's a bad choice because it doesn't seek well, but it plays better than hls when it does play

src/components/deviceprofileBuilder.ts
hawken93: Audio doesn't do passthrough audio

SteveDinn:
Hey @hawken93. I'm just checking in to see if this PR will make it into master any time in the near future. 99% of my video media is x264 + AC3/EAC3 in an MKV container, exactly what you stated in the description, so this PR would be the last thing to make the nightly CC client my go-to.

Thanks for all your work.

hawken93:
Hi @SteveDinn, the misc fix PR goes first, but the important part to figure out here is that without doing anything special we can either
a: make directplay brilliant, hls trashy
b: no directplay, hls works but always low bitrate

The challenge is to get the best of both by:
1. trying to get mkv directplay
2. (if not) trying to get HLS with reduced bitrate..

I can get back to this after the misc fixes are in :)

SteveDinn:
For what it's worth, I have found through experimentation that unchecking the "Allow audio playback that requires transcoding" checkbox in the dashboard user page allows AC3/EAC3 to be passed through. If I have it checked, it will *always* transcode to AAC 2.0.

Screenshot_20210327-191414

nielsvanvelzen:
Closing as inactive

PR #105, opened 2020-12-29
Fix bitrate check

hawken93:
Turns out the device overestimates the bitrate. The bitrate promise gets resolved before the actual content has been transferred.

I also took the liberty to try improving the algorithm, initially starting at half of the current size.

Finally, the detected bitrate needs to be cappable by the device maximum function, not just as a fallback.

### DIFF
--- a/src/components/jellyfinActions.ts
+++ b/src/components/jellyfinActions.ts
@@ -473,8 +473,8 @@ export function getLiveStream(
  *
  * FYI this API has a 10MB limit.
  *
- * @param byteSize number of bytes to request
- * @returns the bitrate in bits/s
+ * @param {number} byteSize number of bytes to request
+ * @returns {Promise<number>} Promise for the bitrate in bits/s
  */
 export function getDownloadSpeed(byteSize: number): Promise<number> {
     const path = 'Playback/BitrateTest?size=' + byteSize;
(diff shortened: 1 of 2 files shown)


src/components/jellyfinActions.ts
```diff
@@ -422,25 +422,35 @@ export function getDownloadSpeed(byteSize: number): Promise<any> {
...
+            // wait for the entire body to appear
+            return response.blob();
+        })
+        .then(function () {
+            const responseTimeSeconds = (new Date().getTime() - now) / 1000;
+            const bytesPerSecond = byteSize / responseTimeSeconds;
+            const bitrate = Math.round(bytesPerSecond * 8);
+
+            return bitrate;
+        });
```
ferferga, line 436: Any reason why this is not an arrow function?

ferferga (commented):
Just one question, otherwise LGTM, although I didn't test this.

src/components/jellyfinActions.ts
hawken93: I'll make it one :)

hawken93:
So with the refactor PR I was able to sneak in the most important part, but what I'm doing in this PR now is:
* Arrow functions and JSDoc
* Try to improve the algorithm used to find the bitrate limit. I think this should be faster for slow connections and more accurate for fast connections..

PR #106, opened 2020-12-31
Make linting stricter and fix linting

ferferga:
This PR adds dependabot and extra linting settings to the ones added to #99. They are really similar to the ones enforced in Vue and Web

#99 must go first, as it adds extra packages needed for these settings to work, this PR just adds extra settings that are not related to the TS migration.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,9 +1,9 @@
 root = true
 
 [*]
+charset = utf-8
+end_of_line = lf
 indent_style = space
 indent_size = 4
-end_of_line = lf
-charset = utf-8
-trim_trailing_whitespace = true
 insert_final_newline = true
+trim_trailing_whitespace = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,11 +2,8 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 2
+indent_size = 4
 end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
-
-[*.md]
-trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,8 +2,11 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 4
+indent_size = 2
+end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
-end_of_line = lf
+
+[*.md]
+trim_trailing_whitespace = false
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,97 +16,97 @@ module.exports = {
         'plugin:import/warnings',
         'plugin:import/typescript'
     ],
+    overrides: [
+        {
+            env: {
+                browser: true,
+                es6: true,
+                node: false
+            },
+            files: ['.js', '.ts'],
+            globals: {
+                $scope: 'writable',
+                cast: 'readonly',
+                PRODUCTION: 'readonly'
+            }
+        }
+    ],
     plugins: ['prettier', 'promise', 'import', 'jsdoc'],
     root: true,
     rules: {
+        '@typescript-eslint/explicit-function-return-type': 'error',
+        '@typescript-eslint/no-unused-vars': 'error',
+        '@typescript-eslint/prefer-ts-expect-error': 'error',
         curly: 'error',
-        'padding-line-between-statements': [
-            'error',
-            // Always require blank lines after directives (like 'use-strict'), except between directives
-            { blankLine: 'always', next: '*', prev: 'directive' },
-            { blankLine: 'any', prev: 'directive', next: 'directive' },
-            // Always require blank lines after import, except between imports
-            { blankLine: 'always', prev: 'import', next: '*' },
-            { blankLine: 'any', prev: 'import', next: 'import' },
-            // Always require blank lines before and after every sequence of variable declarations and export
-            {
-                blankLine: 'always',
-                prev: '*',
-                next: ['const', 'let', 'var', 'export']
-            },
-            {
-                blankLine: 'always',
-                prev: ['const', 'let', 'var', 'export'],
-                next: '*'
-            },
-            {
-                blankLine: 'any',
-                prev: ['const', 'let', 'var', 'export'],
-                next: ['const', 'let', 'var', 'export']
-            },
-            // Always require blank lines before and after class declaration, if, do/while, switch, try
-            {
-                blankLine: 'always',
-                prev: '*',
-                next: ['if', 'class', 'for', 'do', 'while', 'switch', 'try']
-            },
-            {
-                blankLine: 'always',
-                prev: ['if', 'class', 'for', 'do', 'while', 'switch', 'try'],
-                next: '*'
-            },
-            // Always require blank lines before return statements
-            { blankLine: 'always', prev: '*', next: 'return' }
-        ],
         'import/newline-after-import': 'error',
         'import/order': 'error',
-        'jsdoc/require-hyphen-before-param-description': 'error',
+        'jsdoc/check-indentation': 'error',
+        'jsdoc/check-param-names': 'error',
+        'jsdoc/check-property-names': 'error',
+        'jsdoc/check-syntax': 'error',
+        'jsdoc/check-tag-names': 'error',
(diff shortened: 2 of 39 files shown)


src/components/jellyfinActions.ts
```diff
@@ -242,7 +242,7 @@ function setRandomUserBackdrop(): Promise<void> {
             // not everyone will want to see adult backdrops rotating on their TV.
             MaxOfficialRating: 'PG-13'
         }
-    }).then(function (result: BaseItemDtoQueryResult) {
+    }).then((result: BaseItemDtoQueryResult) => {
```
hawken93, line 245: Can you fix the new lint errors as well? Ideally PRs that fail checks should not be merged, hence I at least expect the new linting checks to also be corrected for by this PR. If you have time, the other linting checks as well.

In this instance this function should be rewritten to await syntax.

src/components/jellyfinActions.ts
```diff
@@ -381,11 +381,21 @@ export function play($scope: GlobalScope): void {
...
+/**
+ * @param item
+ * @param maxBitrate
+ * @param deviceProfile
+ * @param startPosition
+ * @param mediaSourceId
+ * @param audioStreamIndex
+ * @param subtitleStreamIndex
+ * @param liveStreamId
+ */
```
hawken93, line 398: Don't check this in without completing the JSDoc please, it's not good to add this just to squelsh the lint error/warning, also the one just below.

src/components/jellyfinActions.ts
```diff
@@ -506,12 +526,12 @@ export function getDownloadSpeed(byteSize: number): Promise<number> {
  */
 export function detectBitrate(): Promise<number> {
```
hawken93, line 527: For this function my PR changes to await syntax. So no changes necessary just in this instance.

src/components/maincontroller.ts
```diff
@@ -48,6 +48,9 @@ let broadcastToServer = new Date();
 
 let hasReportedCapabilities = false;
 
+/**
+ *
+ */
```
hawken93, line 53: jsdoc needs to be completed, same for all the following ones.

hawken93 (changes requested):
Looks good, except my annoyance that I'd like the changes to go with the linting so that we're not buried deep in  lint warnings.

Even before this there are some 260 linter warnings/errors and this will increase the errors even more. It will decrease the warnings some due to incomplete JSDoc but I'd rather have linter warnings than empty comment blocks. I think it's a bad thing that eslint even autofixes to empty comment blocks and params without types or descriptions, these should all be warnings too, but they are not.

In such a situation, new code may introduce more linter remarks, not less, because the linter loses its functionality as a mechanism to keep the code clean.


Also, if you know any checks to make sure public class members in typescript also have documentation, then I think this should be added as well.

I understand that you may not be able to make good jsdoc comments, but in this regard we can prioritize not having any errors, and I'll fix jsdoc that requires knowing the code over time.

ferferga:
@hawken93 will do once we finish all the huge refactoring you did. Didn't want to mess with that until everything was in place 👍🏻.

ferferga:
Using all the identation settings from Vue and fixed all the linting errors (warnings from JSDoc are still there)

Most of the remaining earnings can be fixed as soon as we have proper typings with axios client.

Marking as ready to review.

.prettierrc
```diff
@@ -1,6 +1,5 @@
 {
-    "singleQuote": true,
-    "trailingComma": "none",
-    "semi": true,
-    "tabWidth": 4
+  "semi": true,
+  "singleQuote": true,
+  "trailingComma": "none"
```
Artiume, line 4: did we mean to lose tabWidth?

src/components/codecSupportHelper.ts
```diff
@@ -216,9 +221,9 @@ export function getSupportedMP4AudioCodecs(): Array<string> {
  * @returns Supported HLS video codecs.
  */
 export function getSupportedHLSVideoCodecs(): Array<string> {
-    // Currently the server does not support fmp4 which is required
-    // by the HLS spec for streaming H.265 video.
-    return ['h264'];
+  // Currently the server does not support fmp4 which is required
+  // by the HLS spec for streaming H.265 video.
```
Artiume, line 225: is this still true?

.prettierrc
ferferga: The defaults of Prettier are 2 (so don't need to be explicit about this parameter)

2 tabs is more consistent to what we use in other projects.

It's also enforced in .editorconfig so IDEs are aware of this

src/components/codecSupportHelper.ts
ferferga: Don't know as I did the tests just to check that nothing broke while changing between promises to async/await syntax (which was all that was needed to fix the linting and modernize the code according to our other projects)

This is outside the scope of this PR and should be determined later imo.

.prettierrc
hawken93: As we chatted about on matrix, I would prefer to keep 4, although I would be able to accept 2 space indent.

.eslintrc.js
```diff
@@ -1,26 +1,90 @@
 module.exports = {
     root: true,
-    plugins: ['@typescript-eslint', 'promise', 'import', 'jsdoc'],
     env: {
         node: true,
-        es6: true,
-        browser: true
+        browser: true,
+        es6: true
```
hawken93, line 6: There are some reorderings that are still not alphabetical, maybe sort by alphabet?

src/components/documentManager.ts
```diff
@@ -129,65 +132,65 @@ export abstract class DocumentManager {
...
-                } else if (item?.UserData?.UnplayedItemCount) {
-                    this.setPlayedIndicator(item?.UserData?.UnplayedItemCount);
-                } else {
-                    this.setPlayedIndicator(false);
-                }
+        const urls = [
+            await this.getWaitingBackdropUrl(item),
+            await this.getPrimaryImageUrl(item),
+            await this.getLogoUrl(item)
+        ];
```
hawken93, line 151: I'd like to keep the same behavior here, I did have a shot at it in https://github.com/jellyfin/jellyfin-chromecast/pull/190/files#diff-f7d8d074ad9f2f33a0ea4c9774f0aca90b80d72e4e8dba4f79dcdfc9283228d6R144-R183

Basically we switch to async/await but keep Promise.all() around.

hawken93: It's network requests so doing 3 in parallell is much better than 3 in serial..

src/components/jellyfinActions.ts
```diff
@@ -292,6 +306,16 @@ export function getPlaybackInfo(
...
+/**
+ * @param item
+ * @param playSessionId
+ * @param maxBitrate
+ * @param deviceProfile
+ * @param startPosition
+ * @param mediaSource
+ * @param audioStreamIndex
+ * @param subtitleStreamIndex
+ */
```
hawken93, line 318: maybe remove this too, then me or someone who actually documents the function can take away the warning

src/components/jellyfinActions.ts
hawken93: There should now be a lint rule to warn on empty blocks as well, so this should still produce a warning

src/components/maincontroller.ts
hawken93: There should now be a lint rule to warn on empty blocks as well, so this should still produce a warning

.prettierrc
hawken93: It's now 4 spaces, ty :)

.editorconfig
```diff
@@ -3,7 +3,7 @@ root = true
 [*]
 indent_style = space
 indent_size = 4
+end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
```
hawken93, line 9: maybe sort this block as well?

src/components/documentManager.ts
hawken93: Yey :)

src/components/codecSupportHelper.ts
hawken93: As far as I can see, this file does not have any changes relating to this code anymore

hawken93 (approved):

hawken93:
:+1:

PR #108, opened 2021-01-01
WIP: Fix queue being bugged

hawken93:
Finish refactoring window.playlist over to PlaybackManager, also made PlaybackManager a globally reachable singleton so it doesn't need to be passed around anymore.


hawken93:
WIP: Priority to #93

PR #109, opened 2021-01-01
Remove generated api from source

ferferga:
Removed the generated api from source and uses the axios package instead


hawken93 (commented):
Not much to say about the code changes. They look fine.

I like the import style a lot, I find it much easier when everything is imported from one package. However, I heard some talk on matrix about this causing the final size to be very large. This worries me a bit because typescript definitions aren't supposed to generate any code. I'm wondering if jellyfin should release a `@types` style package just for this purpose instead. I think it's important to be careful about this because especially gen1 really doesn't appreciate big and heavy amounts of javascript and all the clients are already knees deep in heavy scripting thanks to inefficient HLS in MPL.

I haven't tested this myself yet and I would have to see that it does still work. I'll be happy to see all the files go because they really get in the way of grepping the source.

thornbill:
The license of the axios client is currently GPL v3 only. I believe that would prevent its usage here since this project is GPL v2.

ferferga:
@thornbill I've been looking for what you mention here. I can't find any info anywhere about GPL2 anywhere, neither in the runtime-generated client (https://github.com/anttiviljami/openapi-client-axios) or axios itself.

The most relevant information about this that I could find is the one located at [OpenAPITools/openapi-generator](https://github.com/OpenAPITools/openapi-generator) [readme's](https://github.com/OpenAPITools/openapi-generator#34---license-information-on-generated-code). It's not very clear to me if they meant something else, as the links of that section are all broken, but they point to the license of the generator (Apache 2). However, they clearly state the following:

```
When code is generated from this project, it shall be considered AS IS and owned by the user of the software. There are no warranties--expressed or implied--for generated code. You can do what you wish with it, and once generated, the code is your responsibility and subject to the licensing terms that you deem appropriate.
```

Even with the broken links, I think it's clear that we can license it as we wish. This leads us to the discussion of what license to use. I think **MIT** is the de-facto choice, so it's in par with the raw Javascript apiclient we have and we don't lock anyone out, even commercial apps.

ferferga:
@hawken93 With the new package located at https://github.com/jellyfin/jellyfin-client-axios, I can confirm that everything is treeshaked correctly and the bundle size it's the same as without the package (as types are stripped from runtime)

Will rebase this branch once the packages are published to npm

YouKnowBlom:
This looks good to me. The generated client is now licensed MIT which should be compatible. Is there anything preventing this from being merged? @thornbill @ferferga

thornbill:
This was superseded by #400

PR #111, opened 2021-01-01
Add basic translation support

### LINKED ISSUE #41: Translate Chromecast implementation
I can't find a way to translate the interface for the Chromecast since it appears in English when launching from a mobile client to the TV.

ferferga:
The implementation of this is a bit hacky, but it works, at least in Chrome 

**(not tested in actual Chromecast device, would be good to have some input to check if Chromecast is picking the user's language automatically)**

Fixes #41

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,4 +1,5 @@
 import { RepeatMode } from './api/generated/models/repeat-mode';
+import i18n from './i18n';
 import './components/maincontroller';
 import './css/jellyfin.css';
 
@@ -19,3 +20,13 @@ window.mediaElement = document.getElementById('video-player');
 window.playlist = [];
 window.currentPlaylistIndex = -1;
 window.repeatMode = RepeatMode.RepeatNone;
+
+window.i18n = i18n;
+
+(document.getElementById(
+    'readyToCast'
+) as HTMLElement).innerText = window.i18n.t('welcome.readyToCast');
+
+(document.getElementById(
+    'selectMedia'
+) as HTMLElement).innerText = window.i18n.t('welcome.selectMedia');
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -1,0 +1,6 @@
+{
+    "welcome": {
+        "readyToCast": "Ready to Cast",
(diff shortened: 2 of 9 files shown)


ferferga:
@jellyfin/web Anyone has an idea on how we can do this using HTML interpolation [as done in web](https://github.com/jellyfin/jellyfin-web/blob/master/src/components/libraryoptionseditor/libraryoptionseditor.template.html#L50) here? So we don't need to use the ``getElementById`` hack.

nielsvanvelzen:
You could use a small template library like Handlebars for that.

src/locales/en.json
```diff
@@ -0,0 +1,6 @@
+{
+    "welcome": {
```
hawken93, line 2: Maybe change this for `waiting` because that's the screen that the code refers to (appstatus = waiting)

src/app.ts
```diff
@@ -19,3 +20,13 @@ window.mediaElement = document.getElementById('video-player');
 window.playlist = [];
 window.currentPlaylistIndex = -1;
 window.repeatMode = RepeatMode.RepeatNone;
+
+window.i18n = i18n;
+
+(document.getElementById(
```
hawken93, line 26: DocumentManager could handle this post-pr :)

webpack.config.js
```diff
@@ -25,6 +26,9 @@ const config = {
             hash: false,
             favicon: 'favicon.ico'
         }),
+        new CopyWebpackPlugin({
```
hawken93, line 29: If possible, can webpack pull it in the normal way?

src/app.ts
hawken93: Maybe with a updateLocale function that we can call if we have to change the locale while the app is running?

YouKnowBlom:
I think this will have to be discussed a bit more before we settle on a solution, since from what I can tell there have been multiple presented. This is a good start but I am a little hesitant of approving it right now without a clear solution in sight.

castillofrancodamian:
This forgot?

nielsvanvelzen:
Closing as inactive

PR #115, opened 2021-01-16
Make chromecast stable work on 10.7

hawken93:
- backports the fix that makes it work on chromium
- fixes errors relating to 3 api endpoints

essentially this code goes in the trash when nightly gets to the point where it replaces stable.


ferferga (approved):
Assuming it was tested, this LGTM

PR #117, opened 2021-03-10
Switch to npm

ferferga:
Changes this project package management from ``yarn`` to ``npm``, as it's being done in all the organization repositories.

Major reasons for switching:

* All the yarn's advantages over npm are non-existent today
* Yarn introduced some breaking changes in 2.x that caused controversies and avoid us to upgrade directly, which means we are stuck in a version in maintenance mode. So far, all the npm major updates could be updated like a charm, and it's likely they will commit to it. The same can't be said about yarn.
* Npm comes with the node installation, so it's frictionless for the user
* Npm is more mainstream and all the package maintainers add instructions with how to use it to install their packages. The same doesn't apply to yarn. Although adapting the commands to yarn is trivial, for new contributors it might cause some friction.

## Extra

* Added dependabot for dependency updates in GitHub Actions


src/components/commandHandler.ts
```diff
@@ -24,7 +24,7 @@ import { playbackManager } from './playbackManager';
 import { DocumentManager } from './documentManager';
 
 export abstract class CommandHandler {
-    private static playerManager: cast.framework.PlayerManager;
+    private static playerManager: framework.PlayerManager;
```
hawken93, line 27: just a question here, I don't fully understand why "cast" is implicit here..

hawken93 (approved):
I didn't see anything that I wanted to do differently. LGTM :)

src/components/commandHandler.ts
ferferga: Found [this](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/28190#issuecomment-414637374) issue, basically the typings are not exported in the index, but in framework.

Shouldn't affect runtime though as it's all TypeScript abstraction as usual :)

Good catch!

ferferga:
I was able to test this today on an ATV device and I had no issues with the switch, so I'm merging this :)

PR #190, opened 2021-05-13
Refactoring

hawken93:
This one contains commits from misc-fixes that aren't supposed to update any behavior (with one small exception). This will clean up misc-fixes if it gets merged first.

The extend method refactoring does change the behavior a bit, since we cannot get rid of it without changing the way the code works.

### DIFF
--- a/src/components/jellyfinActions.ts
+++ b/src/components/jellyfinActions.ts
@@ -409,7 +409,7 @@ export async function detectBitrate(): Promise<number> {
  */
 export function stopActiveEncodings($scope: GlobalScope): Promise<void> {
     const options = {
-        deviceId: window.deviceInfo.deviceId,
+        deviceId: JellyfinApi.deviceId,
         PlaySessionId: undefined
     };
 
--- a/src/components/jellyfinActions.ts
+++ b/src/components/jellyfinActions.ts
@@ -230,7 +230,7 @@ export function play($scope: GlobalScope): void {
         DocumentManager.getAppStatus() == 'audio'
     ) {
         setTimeout(() => {
-            window.mediaManager.play();
+            window.playerManager.play();
 
             if ($scope.mediaType == 'Audio') {
                 DocumentManager.setAppStatus('audio');
--- a/src/components/jellyfinActions.ts
+++ b/src/components/jellyfinActions.ts
@@ -1,7 +1,6 @@
 import {
     getSenderReportingData,
     resetPlaybackScope,
-    extend,
     broadcastToMessageBus
 } from '../helpers';
 
@@ -199,12 +198,22 @@ export function pingTranscoder(
  */
 export function load(
     $scope: GlobalScope,
-    customData: PlaybackProgressInfo,
+    customData: any,
     serverItem: BaseItemDto
 ): void {
     resetPlaybackScope($scope);
 
-    extend($scope, customData);
+    // These are set up in maincontroller.createMediaInformation
+    $scope.playSessionId = customData.playSessionId;
+    $scope.audioStreamIndex = customData.audioStreamIndex;
+    $scope.subtitleStreamIndex = customData.subtitleStreamIndex;
+    $scope.startPositionTicks = customData.startPositionTicks;
+    $scope.canSeek = customData.canSeek;
+    $scope.itemId = customData.itemId;
+    $scope.liveStreamId = customData.liveStreamId;
+    $scope.mediaSourceId = customData.mediaSourceId;
+    $scope.playMethod = customData.playMethod;
+    $scope.runtimeTicks = customData.runtimeTicks;
 
     $scope.item = serverItem;
 
--- a/src/components/jellyfinApi.ts
+++ b/src/components/jellyfinApi.ts
@@ -11,17 +11,46 @@ export abstract class JellyfinApi {
     // Address of server
     public static serverAddress: string | null = null;
 
+    // device name
+    public static deviceName = 'Google Cast';
+
+    // unique id
+    public static deviceId = '';
+
+    // version
+    public static versionNumber = RECEIVERVERSION;
+
     public static setServerInfo(
         userId: string,
         accessToken: string,
-        serverAddress: string
+        serverAddress: string,
+        receiverName: string
     ): void {
         console.debug(
-            `JellyfinApi.setServerInfo: user:${userId}, token:${accessToken}, server:${serverAddress}`
+            `JellyfinApi.setServerInfo: user:${userId}, token:${accessToken}, ` +
+                `server:${serverAddress}, name:${receiverName}`
         );
         this.userId = userId;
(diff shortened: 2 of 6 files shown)


src/components/jellyfinActions.ts
```diff
@@ -199,12 +198,22 @@ export function pingTranscoder(
  */
 export function load(
     $scope: GlobalScope,
-    customData: PlaybackProgressInfo,
+    customData: any,
```
hawken93, line 201: sorry in advance..
as mentioned earlier, extend was able to put properties on customData that did not even belong in the PlaybackProgressInfo type, so in reality customData is a bit of an extended type. I think we should get rid of $scope and clean up the types from there.

src/components/jellyfinApi.ts
```diff
@@ -11,28 +11,55 @@ export abstract class JellyfinApi {
...
+
+        if (receiverName) {
+            this.deviceName = receiverName;
+            // deviceId just needs to be unique-ish
+            this.deviceId = btoa(receiverName);
+        } else {
+            const senders =
+                cast.framework.CastReceiverContext.getInstance().getSenders();
+
+            this.deviceName = 'Google Cast';
```
ferferga, line 48: Non blocking, but there's any chance this can be the real device name? i.e Chromecast Ultra, JBL, Chromecast Audio...

neopc10: I doubt it especially since groups are a thing, can be generated on the fly, and the device can be switched from the Google Home platform.

ferferga: Yeah but that shouldn't be a problem when the API exposes that?

The question is if the Cast API does.

Other wrappers like pyChromecast from Home Assistant does for example, but all of them are "clients", not "servers" like in this case.

hawken93: components/castDevices.ts uses codecs to figure out what type of chromecast we are dealing with, but not to the most precise degree. However, this is mostly a placeholder.

You cannot really figure out the identity of the chromecast from within this javascript, so what is done is that the cast client passes on the name that the device should take on `if (receiverName)`.

I think it's intentional that google does not expose unique identifiers to let developers recognize the same cast device.. https://stackoverflow.com/questions/23659927/is-there-a-way-to-get-a-unique-id-to-identify-a-chromecast-device


See also this that more directly answers the question: https://stackoverflow.com/questions/41706028/how-to-detect-chromecast-model-in-receiver

It's provided on the initiator side (your phone). So we have to get the ID sent to us.

ferferga: @hawken93 I see, thank you for the detailed explanation 👍

dkanada:
@hawken93 just making sure, this one should be merged before your other open pull requests? Did you have more changes to make or can it be removed from draft status?

src/components/maincontroller.ts
```diff
@@ -251,38 +259,30 @@ export function processMessage(data: any): void {
...
+    data.options = data.options || {};
+
     // Items will have properties - Id, Name, Type, MediaType, IsFolder
 
     JellyfinApi.setServerInfo(
         data.userId,
         data.accessToken,
-        data.serverAddress
+        data.serverAddress,
+        data.receiverName || ''
```
dkanada, line 270: Should probably handle this in the method.

src/components/maincontroller.ts
```diff
@@ -170,34 +172,40 @@ function defaultOnStop(): void {
...
 
-mgr.addEventListener(cast.framework.events.EventType.ENDED, () => {
-    // Ignore
-    if ($scope.isChangingStream) {
-        return;
-    }
+window.playerManager.addEventListener(
+    cast.framework.events.EventType.ENDED,
+    (): void => {
+        // Ignore
```
dkanada, line 187: A more descriptive comment would be useful here.

src/components/maincontroller.ts
```diff
@@ -831,7 +834,9 @@ export function createMediaInformation(
         );
     }
 
-    mediaInfo.customData.startPositionTicks = streamInfo.startPosition || 0;
+    // If the client actually sets startPosition:
+    // if(streamInfo.startPosition)
+    //     mediaInfo.customData.startPositionTicks = streamInfo.startPosition
```
dkanada, line 839: Dead code shouldn't exist in a repository in my opinion, but it you'd rather leave it that's fine.

src/components/jellyfinApi.ts
```diff
@@ -11,28 +11,55 @@ export abstract class JellyfinApi {
...
     // create the necessary headers for authentication
     private static getSecurityHeaders(): Dictionary<string> {
         // TODO throw error if this fails
 
         let auth =
-            `Emby Client="Chromecast", ` +
-            `Device="${window.deviceInfo.deviceName}", ` +
-            `DeviceId="${window.deviceInfo.deviceId}", ` +
-            `Version="${window.deviceInfo.versionNumber}"`;
+            `Emby Client="Chromecast", Device="${this.deviceName}", ` +
```
dkanada, line 61: Relatively certain we support `Jellyfin Client` but we should confirm in chat.

src/components/maincontroller.ts
```diff
@@ -154,14 +150,20 @@ window.addEventListener('beforeunload', () => {
...
 
-mgr.addEventListener(cast.framework.events.EventType.PAUSE, (): void => {
-    reportPlaybackProgress($scope, getReportingParams($scope));
-});
+window.playerManager.addEventListener(
+    cast.framework.events.EventType.PAUSE,
+    (): void => {
+        reportPlaybackProgress($scope, getReportingParams($scope));
+    }
+);
```
dkanada, line 166: I'm guessing this was done to reduce the line length but the previous style was much more readable in my opinion.

src/components/maincontroller.ts
hawken93: I think streamInfo.startPosition was an unknown variable to me, and I think the client would have passed this position, but it's a duplicate property imo. I assumed maybe someone knew something more about it, but if it actually doesn't do anything then it should be removed. :)

src/components/maincontroller.ts
hawken93: It was previously `mgr.addEventListener` which was refactored to `window.playerManger`. prettier is unfortunately the master of the line breaks here :|

src/components/maincontroller.ts
hawken93: If nobody has any relation to streamInfo.startPosition, and nothing is found in jellyfin-web, I think it should be deleted. :)

YouKnowBlom (changes requested):
I think the open conversations are relevant but other than that LGTM!

YouKnowBlom (approved):
LGTM! My changes are unreviewed but as they're very minor, I'll consider them OK.

PR #320, opened 2021-07-22
Refactor credentialManager to use new interface instead of Configuration

YouKnowBlom:
The previously used interface `Configuration` did not seem quite right for the purpose. I tried looking for an alternative in the generated api client with no luck. Creating our own interface should allow for more flexibility. Although not a priority right now, this is one step towards support for multiple servers.


dkanada:
We are trying to move from API clients to a more useful SDK approach that provides useful functionality alongside the generated code, so anything of that nature should be added to whatever repository we eventually back for JavaScript.

dkanada (approved):

PR #338, opened 2021-07-30
Add tests for jellyfinApi

YouKnowBlom:
Change  `JellyfinApi.setServerInfo` parameters to accept null.
Add `--silent` to test command. This will silence output printed via `console`
Add check to remove leading slashes from `path` parameter in `JellyfinApi.createImageUrl`.
Add basic tests to `JellyfinApi`.
Change test environment to `jsdom` as that provides a more browser-like test environment.
Define `RECEIVERVERSION` globally for tests. Since we're not running the code through webpack, this variable will never be replaced and jest will get angry.

### DIFF
--- a/src/components/jellyfinApi.ts
+++ b/src/components/jellyfinApi.ts
@@ -3,13 +3,13 @@ import { Dictionary } from '~/types/global';
 
 export abstract class JellyfinApi {
     // userId that we are connecting as currently
-    public static userId: string | null = null;
+    public static userId: string | undefined;
 
     // Security token to prove authentication
-    public static accessToken: string | null = null;
+    public static accessToken: string | undefined;
 
     // Address of server
-    public static serverAddress: string | null = null;
+    public static serverAddress: string | undefined;
 
     // device name
     public static deviceName = 'Google Cast';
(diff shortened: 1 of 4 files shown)


src/components/jellyfinApi.ts
```diff
@@ -21,9 +21,9 @@ export abstract class JellyfinApi {
     public static versionNumber = RECEIVERVERSION;
 
     public static setServerInfo(
-        userId: string,
-        accessToken: string,
-        serverAddress: string,
+        userId: string | null,
+        accessToken: string | null,
+        serverAddress: string | null,
```
dkanada, line 26: Why not just do something like this? I'm not 100% familiar with default values in typescript so maybe the behavior is different but they seem nicer.

```
userId: string = null
```

YouKnowBlom: The behavior is technically different but ultimately achieve the same thing, more or less. the original approach allows null to be passed while the new one would also allow it to be passed, while also defaulting to null. I do see the benefit of having to pass less parameters when only the first or second are needed. I don't have a strong opinion about this so if you feel that the new approach would be better then I'll switch. However it would still have to have null as a possible type, or else the TS compiler will yell at us for assigning null to a string type:

```
userId: string | null = null
```

dkanada: This sounds good enough then. I'm actually just surprised TypeScript doesn't have a better syntax for handling nullability than this to be honest.

nielsvanvelzen (approved):

PR #347, opened 2021-10-30
Migrate from file-loader webpack asset modules

rhysmdnz:
This fixes images such as the jellyfin logo failing to load


printfn:
Is there anything blocking this from getting merged?

nielsvanvelzen (changes requested):
I think you can `npm rm file-loader` now.

rhysmdnz:
It seems so. Removed :)

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks for the wait - as you might have noticed the development on the cast client stalled so I can't say anything on when this would be published with a new version. I believe the master branch is deployed to the unstable instance though.

PR #470, opened 2023-09-28
fix aac codecprofile condition

Sky-High:
- remove conditions for IsSecondaryAudio, as it is also defined for "all codecs"
- in stead, restore condition AudioChannels which disappeared in commit #4e3dc81, pr#74 probably because of a copy/paste glitch. See https://github.com/jellyfin/jellyfin-chromecast/blob/96a8865271d4ecf640fd9a970b21ed0127c02341/src/components/deviceprofilebuilder.js#L141
- the latter restore also removes some erroneous unnecessary transcoding.


nielsvanvelzen:
The "IsSecondaryAudio" property is true for all embedded audio tracks that are not the first one. The reason this exists is because some (all?) browser cannot switch audio tracks for playing media so you need the server to change the audio tracks to play a secondary one.

See also, the implementation on the server:
https://github.com/jellyfin/jellyfin/blob/626f474faaa1f0511f6241fe95a9bb288e50343b/MediaBrowser.Model/Dto/MediaSourceInfo.cs#L231-L248

The test in jellyfin-web:
https://github.com/jellyfin/jellyfin-web/blob/128605e3fcde13db9f62b2e46b97610c356d758b/src/scripts/browserDeviceProfile.js#L376-L391

nielsvanvelzen (approved):

PR #476, opened 2023-10-02
Replace Webpack with Vite

nielsvanvelzen:
This PR removed Webpack in favor of Vite. The configuration is much simpler, there are way less dependencies, build times improve significantly and we get hot reload for the index.html file.


src/helpers.ts
```diff
@@ -43,7 +43,7 @@ export function getReportingParams(state: PlaybackState): PlaybackProgressInfo {
     return {
         AudioStreamIndex: state.audioStreamIndex,
         CanSeek: state.canSeek,
-        IsMuted: window.volume.muted,
+        IsMuted: window.volume?.muted,
```
thornbill, line 46: This could be undefined, should it be coerced to be a Boolean?

src/helpers.ts
```diff
@@ -55,7 +55,7 @@ export function getReportingParams(state: PlaybackState): PlaybackProgressInfo {
         PositionTicks: Math.round(getCurrentPositionTicks(state)),
         RepeatMode: window.repeatMode,
         SubtitleStreamIndex: state.subtitleStreamIndex,
-        VolumeLevel: Math.round(window.volume.level * 100)
+        VolumeLevel: Math.round(window.volume?.level * 100)
```
thornbill, line 58: Similarly this could be `undefined * 100`?

ferferga (commented):
I believe ts-node, cross-env and source-map-loader are not necessary as well?

nielsvanvelzen:
Yup and I also forgot css-loader

vite.config.ts
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable sort-keys */
+
+import { defineConfig } from 'vite';
+import { version } from './package.json';
+
+export default defineConfig({
+    root: 'src',
+    build: {
+        outDir: '../dist',
```
thornbill, line 9: `../` doesn't seem right here?

nielsvanvelzen: The outDir is relative to the root, our sources are in the "src" folder and we want the dist folder outside of it. Just stupid JavaScript ecosystem things.

nielsvanvelzen: See https://vitejs.dev/config/build-options.html#build-outdir

thornbill (approved):

PR #477, opened 2023-10-02
Remove development build

nielsvanvelzen:
This one simplifies the build (specifically CI) by removing the development build. If you want dev you use `npm run start`, if you want a build you use `npm run build` and get a production build.

Requirement for #476 

_🤞 this doesn't break CI again_

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -14,9 +14,6 @@ jobs:
     build:
         name: Build
         runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
Oh yeah build fails because it runs the workflow from master... well that makes only one way to verify it works....

nielsvanvelzen:
It actually did run the workflow from this branch but I forgot something. I might need some sleep.

PR #506, opened 2023-11-28
Fix sync of volume level with webclient

Sky-High:
The volume level slider of the jellyfin webclient is not synchronized with the chromecast webreceiver volume setting.
This is fixed here.

**Changes**
Add appropriate parentheses.


thornbill (approved):
Nice find!

PR #537, opened 2024-02-29
Expand readme (inspired by server-repo)

tim-vk:
A set-up for a more complete README. I did notice some oddities while writing it though:

- Current `latest` version is 1.0.1 however there is also a version 1.1.0 version available which seems odd. Is this a mistake? (I do think 1.0.1 is newer, but the number is off. Would assume it to be 1.1.1 or 1.2.0).
- The bit about `testing`: We can switch between a stable & unstable version. But no where can I actually find an actual implementation of something that does something with this setting.  The jellyfin-web client seems to use it's own chromecast plugin. So is this repo even used a.t.m.? (if so: can you point me to the right direction so I can figure out stuff further from there?).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<h1 align="center">Jellyfin Chromecast Client</h1>
+<h1 align="center">Jellyfin Chromecast Web Receiver</h1>
 <h3 align="center">Part of the <a href="https://jellyfin.org">Jellyfin Project</a></h3>
 
 <p align="center">
@@ -13,25 +13,34 @@
 </a>
 </p>
 
-Jellyfin chromecast is a receiver used when casting to a google cast capable device. It is automatically used when casting from the jellyfin android client or jellyfin web client.
+Jellyfin Chromecast is a receiver app used when casting to a Google Cast capable device. It is used when casting from the Jellyfin Android client or Jellyfin web client.
+
+### How do I use it?
+
+A `stable` and `unstable` version of this app are already included in `jellyfin-server`. There is no need to seperately install this project. To host your own version (for developing) see `CONTRIBUTING.md`.
 
 ### What does it do?
 
-As soon as you push the "cast" button on your client this application will start on you cast-capable device & handle it from there.
+This is a `web receiver` as defined in the [Google Chromecast architecture](https://developers.google.com/cast/docs/overview).
+
+As soon as you push the "cast" button on your client this application will start on you cast-capable device & handle it from there. 
 
 ### What doesn't it do?
 
-Anything related to your non-cast device (e.g. your phone, browser, other device). Any issues/features related to that: check the respective repository.
+Anything related to your non-cast device (e.g. your phone, browser, other device) or anything about the inclusion of casting for a specific client (e.g. casting from the iOS app).
+
+Any issues/features related to that: check the respective repository.
 
 ### Something not working right?
-First check if the issue is actually google-cast related. So answer the question:
+First check if the issue is actually Google Cast related. So answer the question:
 
-`"Can I reproduce the issue on any other device than a google chromecast?"`
+`"Can I reproduce the issue on any other device than a Google Chromecast?"`
 
-If yes: Then the issue probably lies somewhere else. If no: Open an <a href="https://github.com/jellyfin/jellyfin-chromecast/issues/new/choose">Issue</a> on GitHub.<br/>
+If yes: The issue probably lies somewhere else. 
+If no: Open an <a href="https://github.com/jellyfin/jellyfin-chromecast/issues/new/choose">Issue</a> on GitHub.<br/>
 
 ### Testing
 
-Jellyfin allows switching between a `stable` and `unstable` version of the client. Go the client of your choice and: `user` -> `settings` -> `playback` -> `Google cast version`.
+Jellyfin allows switching between a `stable` and `unstable` version of the client. Go the client of your choice and: `user` -> `settings` -> `playback` -> `Google Cast version`.
 
 Note that this setting is set per-user.
--- a/README.md
+++ b/README.md
@@ -1,3 +1,37 @@
-# Jellyfin Chromecast
+<h1 align="center">Jellyfin Chromecast Client</h1>
+<h3 align="center">Part of the <a href="https://jellyfin.org">Jellyfin Project</a></h3>
 
-Source code for the Jellyfin Chromecast receiver application
+<p align="center">
+<img alt="Logo Banner" src="https://raw.githubusercontent.com/jellyfin/jellyfin-ux/master/branding/SVG/banner-logo-solid.svg?sanitize=true"/>
+<br/>
+<br/>
+<a href="https://github.com/jellyfin/jellyfin">
+<img alt="GPL 2.0 License" src="https://img.shields.io/github/license/jellyfin/jellyfin-chromecast.svg"/>
+</a>
+<a href="https://github.com/jellyfin/jellyfin/releases">
+<img alt="Current Release" src="https://img.shields.io/github/release/jellyfin/jellyfin-chromecast.svg"/>
+</a>
+</p>
+
+Jellyfin chromecast is a receiver used when casting to a google cast capable device. It is automatically used when casting from the jellyfin android client or jellyfin web client.
+
+### What does it do?
+
+As soon as you push the "cast" button on your client this application will start on you cast-capable device & handle it from there.
+
+### What doesn't it do?
+
+Anything related to your non-cast device (e.g. your phone, browser, other device). Any issues/features related to that: check the respective repository.
+
+### Something not working right?
+First check if the issue is actually google-cast related. So answer the question:
+
+`"Can I reproduce the issue on any other device than a google chromecast?"`
+
+If yes: Then the issue probably lies somewhere else. If no: Open an <a href="https://github.com/jellyfin/jellyfin-chromecast/issues/new/choose">Issue</a> on GitHub.<br/>
+
+### Testing
+
+Jellyfin allows switching between a `stable` and `unstable` version of the client. Go the client of your choice and: `user` -> `settings` -> `playback` -> `Google cast version`.
+
+Note that this setting is set per-user.
--- a/README.md
+++ b/README.md
@@ -1,29 +1,29 @@
-<h1 align="center">Jellyfin Chromecast Web Receiver</h1>
+<h1 align="center">Jellyfin Cast Web Receiver</h1>
 <h3 align="center">Part of the <a href="https://jellyfin.org">Jellyfin Project</a></h3>
 
 <p align="center">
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,3 +1,37 @@
-# Jellyfin Chromecast
+<h1 align="center">Jellyfin Chromecast Client</h1>
```
thornbill, line 1: Jellyfin Chromecast Web Receiver or Jellyfin Chromecast Receiver App might be better names?

README.md
```diff
@@ -1,3 +1,37 @@
...
+<a href="https://github.com/jellyfin/jellyfin/releases">
+<img alt="Current Release" src="https://img.shields.io/github/release/jellyfin/jellyfin-chromecast.svg"/>
+</a>
+</p>
+
+Jellyfin chromecast is a receiver used when casting to a google cast capable device. It is automatically used when casting from the jellyfin android client or jellyfin web client.
+
+### What does it do?
+
+As soon as you push the "cast" button on your client this application will start on you cast-capable device & handle it from there.
```
thornbill, line 20: It would probably be good to provide this link for reference to the overall casting process/architecture: https://developers.google.com/cast/docs/overview

README.md
```diff
@@ -1,3 +1,37 @@
...
+
+### What doesn't it do?
+
+Anything related to your non-cast device (e.g. your phone, browser, other device). Any issues/features related to that: check the respective repository.
```
thornbill, line 24: It would be good to specifically point out that this is not related to clients supporting casting either. i.e. Don't open issues here about the iOS app not supporting casting.

README.md
```diff
@@ -1,3 +1,37 @@
...
+<br/>
+<a href="https://github.com/jellyfin/jellyfin">
+<img alt="GPL 2.0 License" src="https://img.shields.io/github/license/jellyfin/jellyfin-chromecast.svg"/>
+</a>
+<a href="https://github.com/jellyfin/jellyfin/releases">
+<img alt="Current Release" src="https://img.shields.io/github/release/jellyfin/jellyfin-chromecast.svg"/>
+</a>
+</p>
+
+Jellyfin chromecast is a receiver used when casting to a google cast capable device. It is automatically used when casting from the jellyfin android client or jellyfin web client.
```
thornbill, line 16:
```suggestion
Jellyfin Chromecast is a receiver app used when casting to a Google Cast capable device. It is used when casting from the Jellyfin Android client or Jellyfin web client.
```

It's probably also worth noting that the default instance of the app is hosted by Jellyfin.

README.md
```diff
@@ -1,3 +1,37 @@
...
+### What does it do?
+
+As soon as you push the "cast" button on your client this application will start on you cast-capable device & handle it from there.
+
+### What doesn't it do?
+
+Anything related to your non-cast device (e.g. your phone, browser, other device). Any issues/features related to that: check the respective repository.
+
+### Something not working right?
+First check if the issue is actually google-cast related. So answer the question:
```
thornbill, line 27:
```suggestion
First check if the issue is actually Google Cast related. So answer the question:
```

README.md
```diff
@@ -1,3 +1,37 @@
...
+
+`"Can I reproduce the issue on any other device than a google chromecast?"`
```
thornbill, line 29:
```suggestion
> "Can I reproduce the issue on any other device than a Google Chromecast?"
```

README.md
```diff
@@ -1,3 +1,37 @@
...
+
+If yes: Then the issue probably lies somewhere else. If no: Open an <a href="https://github.com/jellyfin/jellyfin-chromecast/issues/new/choose">Issue</a> on GitHub.<br/>
```
thornbill, line 31:
```suggestion
* If yes: The issue probably lies somewhere else.
* If no: [Open an issue here.](https://github.com/jellyfin/jellyfin-chromecast/issues/new/choose)
```

README.md
tim-vk: I agree `Jellyfin Chromecast Receiver (app)` would be better suited. But not sure if there is a different name that refers to this repo. Setting it to your suggestion now but i'm open to any suggestions.

README.md
tim-vk: I've added a seperate "how do i use it". I did notice that the current contributing is not really up to date anymore. Since that refers to editing `jellyfin-web` if you want to host your own version. That's no longer possible since this mr: https://github.com/jellyfin/jellyfin-meta/issues/45

Though I'm uncertain how the workflow should look like now. But that's not an issue for this readme :) I've referred to the contributing.md.

README.md
```diff
@@ -1,3 +1,46 @@
-# Jellyfin Chromecast
+<h1 align="center">Jellyfin Chromecast Web Receiver</h1>
+<h3 align="center">Part of the <a href="https://jellyfin.org">Jellyfin Project</a></h3>
 
-Source code for the Jellyfin Chromecast receiver application
+<p align="center">
+<img alt="Logo Banner" src="https://raw.githubusercontent.com/jellyfin/jellyfin-ux/master/branding/SVG/banner-logo-solid.svg?sanitize=true"/>
+<br/>
+<br/>
+<a href="https://github.com/jellyfin/jellyfin">
```
nielsvanvelzen, line 8: Should link to this repo

README.md
```diff
@@ -1,3 +1,46 @@
-# Jellyfin Chromecast
+<h1 align="center">Jellyfin Chromecast Web Receiver</h1>
```
nielsvanvelzen, line 1:
```suggestion
<h1 align="center">Jellyfin Cast Web Receiver</h1>
```

Chromecast is renamed to Google Cast or just Cast these days, so let's use that for the name (even though the repository url still uses the chromecast name)

README.md
```diff
@@ -1,3 +1,46 @@
...
+<br/>
+<a href="https://github.com/jellyfin/jellyfin">
+<img alt="GPL 2.0 License" src="https://img.shields.io/github/license/jellyfin/jellyfin-chromecast.svg"/>
+</a>
+<a href="https://github.com/jellyfin/jellyfin/releases">
+<img alt="Current Release" src="https://img.shields.io/github/release/jellyfin/jellyfin-chromecast.svg"/>
+</a>
+</p>
+
+Jellyfin Chromecast is a receiver app used when casting to a Google Cast capable device. It is used when casting from the Jellyfin Android client or Jellyfin web client.
```
nielsvanvelzen, line 16:
```suggestion
The Jellyfin cast web receiver is the frontend used when casting to a Google Cast capable device. It is used by default when casting from the Jellyfin Android app or Jellyfin web client.
```

README.md
```diff
@@ -1,3 +1,46 @@
...
+
+### How do I use it?
+
+A `stable` and `unstable` version of this app are already included in `jellyfin-server`. There is no need to seperately install this project. To host your own version (for developing) see `CONTRIBUTING.md`.
+
+### What does it do?
+
+This is a `web receiver` as defined in the [Google Chromecast architecture](https://developers.google.com/cast/docs/overview).
+
+As soon as you push the "cast" button on your client this application will start on you cast-capable device & handle it from there.
```
nielsvanvelzen, line 26:
```suggestion
As soon as you press the "cast" button on your client this application will start on you cast-capable device and handle playback functionality. 
```

README.md
```diff
@@ -1,3 +1,46 @@
...
+As soon as you push the "cast" button on your client this application will start on you cast-capable device & handle it from there. 
+
+### What doesn't it do?
+
+Anything related to your non-cast device (e.g. your phone, browser, other device) or anything about the inclusion of casting for a specific client (e.g. casting from the iOS app).
+
+Any issues/features related to that: check the respective repository.
+
+### Something not working right?
+First check if the issue is actually Google Cast related. So answer the question:
```
nielsvanvelzen, line 35:
```suggestion
### Something not working right?

First check if the issue is actually Google Cast related. So answer the question:
```

README.md
```diff
@@ -1,3 +1,46 @@
...
+
+`"Can I reproduce the issue on any other device than a Google Chromecast?"`
```
nielsvanvelzen, line 37:
```suggestion
`"Can I reproduce the issue on any other device than a Google Chromecast?"`
```

My Android TV contains Chromecast functionality but it is not a Google Chromecast. Not sure how to rephrase it though.

README.md
```diff
@@ -1,3 +1,46 @@
...
+
+If yes: The issue probably lies somewhere else. 
+If no: Open an <a href="https://github.com/jellyfin/jellyfin-chromecast/issues/new/choose">Issue</a> on GitHub.<br/>
```
nielsvanvelzen, line 40: Use markdown syntax for link

nielsvanvelzen, line 40:
```suggestion
If no: Open an <a href="https://github.com/jellyfin/jellyfin-chromecast/issues/new/choose">Issue</a> on GitHub.
```

README.md
tim-vk: Oof... I see what you mean. Additionally: you can both cast to the android tv and run the android tv jellyfin client. (And only the first one is relevant for this repo). I've changed the wording. But open to other suggestions.

tim-vk:
@nielsvanvelzen : Thanks for the review! I've made some changes accordingly.

Please don't forget the two questions i've put in the description:

- Current latest version is 1.0.1 however there is also a version 1.1.0 version available which seems odd. Is this a mistake? (I do think 1.0.1 is newer, but the number is off. Would assume it to be 1.1.1 or 1.2.0).
- The bit about testing: We can switch between a stable & unstable version. But no where can I actually find an actual implementation of something that does something with this setting. The jellyfin-web client seems to use it's own chromecast plugin. So is this repo even used a.t.m.? (if so: can you point me to the right direction so I can figure out stuff further from there & update the CONTRIBUTING.md accordingly?).

nielsvanvelzen:
Thanks for the reminder about your questions, I skipped the description 😅

> Current latest version is 1.0.1 however there is also a version 1.1.0 version available which seems odd. Is this a mistake? (I do think 1.0.1 is newer, but the number is off. Would assume it to be 1.1.1 or 1.2.0).

I'm honestly not sure what happened as I didn't do anything with this repo when those releases were made. My best guess is that 1.0.1 was incorrectly labelled and should've been 1.1.1. The diff appears to contain three commits: https://github.com/jellyfin/jellyfin-chromecast/compare/v1.1.0...v1.0.1. When I look at the source of the deployed CC app I can see the added comment in embyactions.js so 1.0.1 is definitely the "latest" and deployed version.

> The bit about testing: We can switch between a stable & unstable version. But no where can I actually find an actual implementation of something that does something with this setting. The jellyfin-web client seems to use it's own chromecast plugin. So is this repo even used a.t.m.? (if so: can you point me to the right direction so I can figure out stuff further from there?).

We've made changes for 10.9 that make it possible to add your own receiver. Documentation for that is in jellyfin/jellyfin-meta#45. Note that the actual implementation might differ from the proposal.

nielsvanvelzen (approved):
Awesome, thanks for your work!

PR #622, opened 2024-10-20
Update to ESLint 9

3flex:
I'm no expert here, so there may be some issues in the migration of the config file. The new config file format is very different to the old, and some things weren't clear in the ESLint config migration guide for v9, so it's possible that there are some unintended behaviour changes that have been introduced.

I've verified that the typescript checks are working which is probably the most important thing.


.eslintrc.js
```diff
@@ -1,123 +0,0 @@
...
-            env: {
-                browser: true,
-                es6: true,
-                node: false
-            },
-            files: ['.js', '.ts'],
-            globals: {
-                $scope: 'writable',
-                cast: 'readonly'
-            }
```
3flex, line 29: I didn't understand the reason for including these items in the `globals` block, and a look at the commit history and a few searches online didn't help.

They have not been carried over to the new config.

nielsvanvelzen (approved):

nielsvanvelzen:
The linter seems to run fine, we can always update the config later if any issues with it arise.

PR #627, opened 2024-10-21
Enable bundler module resolution mode

3flex:
Previous "node" option should no longer be used:
https://www.typescriptlang.org/docs/handbook/modules/reference.html#node10-formerly-known-as-node

"bundler" is recommended when using a bundler.


nielsvanvelzen:
This one has a merge conflict now

nielsvanvelzen (approved):

PR #644, opened 2024-10-25
Use TvShowsApiGetEpisodesRequest type for getEpisodes query parameters

3flex:


nielsvanvelzen (approved):

3flex:
Just wondering @nielsvanvelzen would it be a good idea to take advantage of the Jellyfin TypeScript SDK and use the Axios client? It seems some of the implementation in this project could be simplified. I had a look around when taking a look for this change.

nielsvanvelzen:
Yup definitely, migrating fully to the SDK would make everything more type-safe too :)

PR #645, opened 2024-10-25
Use deviceIds enum type where possible

3flex:


nielsvanvelzen (commented):
We should rename the enum to `DeviceIds` (PascalCase) so it won't be confused with regular variables.

nielsvanvelzen (approved):

PR #648, opened 2024-10-26
Remove volume command handlers

3flex:
These have been no-ops for a long time.


3flex:
Closing as these may still be used by the Android client and I don't plan to investigate that yet.

PR #650, opened 2024-10-27
Start using generated Typescript API for API calls

3flex:
This PR:
* Sets up the SDK & API instances
* Converts three playback actions to use the generated API instead of a fetch helper

I wanted to see if this initial implementation was acceptable before continuing.

Tested successfully locally.

### DIFF
--- a/src/components/jellyfinActions.ts
+++ b/src/components/jellyfinActions.ts
@@ -2,8 +2,8 @@ import type {
     BaseItemDto,
     DeviceProfile,
     LiveStreamResponse,
-    MediaInfoApiGetPostedPlaybackInfoRequest,
     MediaSourceInfo,
+    PlaybackInfoDto,
     PlaybackProgressInfo,
     PlayRequest
 } from '@jellyfin/sdk/lib/generated-client';
@@ -260,20 +260,34 @@ export async function getPlaybackInfo(
         return Promise.reject('Item ID not available.');
     }
 
-    const query: MediaInfoApiGetPostedPlaybackInfoRequest = {
-        audioStreamIndex: audioStreamIndex ?? undefined,
-        itemId: item.Id,
-        liveStreamId: liveStreamId ?? undefined,
-        maxStreamingBitrate: maxBitrate,
-        mediaSourceId: mediaSourceId ?? undefined,
-        startTimeTicks: startPosition || 0,
-        subtitleStreamIndex: subtitleStreamIndex ?? undefined,
-        userId: JellyfinApi.userId ?? undefined
+    const query: PlaybackInfoDto = {
+        MaxStreamingBitrate: maxBitrate,
+        StartTimeTicks: startPosition || 0,
+        UserId: JellyfinApi.userId
     };
+
+    if (audioStreamIndex != null) {
+        query.AudioStreamIndex = audioStreamIndex;
+    }
+
+    if (subtitleStreamIndex != null) {
+        query.SubtitleStreamIndex = subtitleStreamIndex;
+    }
+
+    if (mediaSourceId) {
+        query.MediaSourceId = mediaSourceId;
+    }
+
+    if (liveStreamId) {
+        query.LiveStreamId = liveStreamId;
+    }
 
     const response = await getMediaInfoApi(
         JellyfinApi.jellyfinApi
-    ).getPostedPlaybackInfo(query);
+    ).getPostedPlaybackInfo({
+        itemId: item.Id,
+        playbackInfoDto: query
+    });
 
     return response.data;
 }
--- a/src/components/jellyfinActions.ts
+++ b/src/components/jellyfinActions.ts
@@ -6,6 +6,7 @@ import type {
     PlaybackProgressInfo,
     PlayRequest
 } from '@jellyfin/sdk/lib/generated-client';
+import { getPlaystateApi } from '@jellyfin/sdk/lib/utils/api';
 import { getSenderReportingData, broadcastToMessageBus } from '../helpers';
 import { AppStatus } from '../types/appStatus';
 import { JellyfinApi } from './jellyfinApi';
@@ -91,7 +92,7 @@ export function reportPlaybackStart(
  * @param broadcastEventName - name of event to send to the cast sender
  * @returns Promise for the http request
  */
-export function reportPlaybackProgress(
+export async function reportPlaybackProgress(
     state: PlaybackState,
     reportingParams: PlaybackProgressInfo,
     reportToServer = true,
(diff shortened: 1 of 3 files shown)


src/components/jellyfinActions.ts
```diff
@@ -76,10 +78,8 @@ export function reportPlaybackStart(
 
     restartPingInterval(reportingParams);
 
-    return JellyfinApi.authAjax('Sessions/Playing', {
-        contentType: 'application/json',
-        data: JSON.stringify(reportingParams),
-        type: 'POST'
+    await getPlaystateApi(JellyfinApi.jellyfinApi).reportPlaybackStart({
```
thornbill, line 81: Why do these not return anything now?

3flex: My understanding is that there's an implicit return, so there's no need to include one since the caller doesn't read anything from the `void` return type. I can add an explicit one if that's preferred?

I'm JS beginner level so I apologise for any basic mistakes - I'm coming from a Kotlin background where there's no need to return from a function with `Unit` (`void`-equivalent) type.

src/components/jellyfinActions.ts
```diff
@@ -239,59 +237,45 @@ export function play(state: PlaybackState): void {
...
     item: BaseItemDto,
     maxBitrate: number,
-    deviceProfile: DeviceProfile,
     startPosition: number,
     mediaSourceId: string,
     audioStreamIndex: number,
     subtitleStreamIndex: number,
     liveStreamId: string | null = null
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
 ): Promise<any> {
```
nielsvanvelzen, line 256:
```suggestion
): Promise<PlaybackInfoResponse> {
```

Could safely use the response type now here

src/components/jellyfinActions.ts
thornbill: It isn't necessary to explicitly return but it does seem odd to use async/await in this way when you could just return the sdk call.

nielsvanvelzen: I believe these specific API calls have no response (204) so it would be void anyway

src/components/jellyfinActions.ts
3flex: I would have but it has knock on effects at the call site so some refactoring would be needed. I wanted to avoid doing that in this PR but I can if it's preferred.

nielsvanvelzen: Fair enough, let's keep it separate then

src/components/jellyfinActions.ts
3flex: [`reportPlaybackStart`](https://github.com/jellyfin/jellyfin-sdk-typescript/blob/v0.10.0/src/generated-client/api/playstate-api.ts#L1181) and [`reportPlaybackProgress`](https://github.com/jellyfin/jellyfin-sdk-typescript/blob/v0.10.0/src/generated-client/api/playstate-api.ts#L1169) both return `void`, but in the generated typings they return `Promise<import("axios").AxiosResponse<void, any>>`.

3flex:
I was wondering about this but it's timely with the JF 10.10 release - does the Cast app need to support a range of JF versions? The TS SDK seems to support a specific major version (currently 10.9 but it's being updated to 10.10) so if the code transitions to use the TS SDK it will generally support whatever version of JF that the TS SDK supports.

There's currently a bit of code that handles different JF versions, if this is cut over to use the TS SDK then backwards compatibility may be broken:

https://github.com/jellyfin/jellyfin-chromecast/blob/22099bfd45cec11d743c3978181d7161aed9e6b4/src/components/jellyfinActions.ts#L167-L178

This may impact the meaning of the "stable" release - if the stable release used the TS SDK v0.10.0 which supports JF 10.9 then it may not work in JF 10.10. Updating the TS SDK in "stable" to a release that supports 10.10 may break Cast in 10.9.

thornbill:
We generally avoid breaking API changes. The general consensus is the API needs a total overhaul and that will coincide with the v11 release. Until then the API should remain as stable as possible.

src/components/jellyfinActions.ts
```diff
@@ -239,59 +237,45 @@ export function play(state: PlaybackState): void {
...
  * @param audioStreamIndex - audioStreamIndex
  * @param subtitleStreamIndex - subtitleStreamIndex
  * @param liveStreamId - liveStreamId
  * @returns promise
  */
-export function getPlaybackInfo(
+export async function getPlaybackInfo(
     item: BaseItemDto,
     maxBitrate: number,
-    deviceProfile: DeviceProfile,
```
3flex, line 253: I just realised removing `deviceProfile` is incorrect. I was confused because I couldn't see a parameter with this type on `MediaInfoApiGetPostedPlaybackInfoRequest`.

It's there on the API though and I see it should actually be passed through with `playbackInfoDto`. And using `PlaybackInfoDto` instead of `MediaInfoApiGetPostedPlaybackInfoRequest` here will make it easier to use the generated API.

I'm assuming it would be best to use the types from the SDK models in client code instead of using the types defined for the API itself.

nielsvanvelzen (approved):

src/components/jellyfinActions.ts
thornbill: This is essentially the same as `return await asyncFn()` vs `return asyncFn()` which has some slight performance implications: https://stackoverflow.com/questions/38708550/difference-between-return-await-promise-and-return-promise

In my opinion returning the call is also just generally more explicit about the intent of the function.

The typing is correct for the functions `Promise<import("axios").AxiosResponse<void, any>>`:
* `Promise`: all async functions implicitly return a Promise
* `AxiosResponse<void, any>`: this is the response of an Axios request that returns `void`

3flex: I'm still missing something.

If I return the call (with or without async/await):
```typescript
    return await getPlaystateApi(JellyfinApi.jellyfinApi).reportPlaybackProgress({
        playbackProgressInfo: reportingParams
    });
```

Then the function return type has to be `Promise<AxiosResponse<void>>`, meaning the `AxiosReponse` has to be handled at the call site. Is that the right way to go? I'd have thought that `AxiosResponse` should be handled in the API module and pass the DTO back to the caller instead (similar to what I've done with `getPlaybackInfo` which returns `Promise<any>` but will be corrected later to return `Promise<PlaybackInfoResponse>`).

thornbill: Having metadata about the request could be useful in some cases which is probably why the generator returns AxiosResponse by default. Since you don't need it this is probably fine. I do still think the code is a bit ambiguous to read but not the end of the world.

PR #657, opened 2024-10-29
Enable no-unnecessary-type-assertion ts eslint rule

3flex:
This is enabled by default in the recommended type checked rule set.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -59,6 +59,7 @@ export default [
         rules: {
             '@typescript-eslint/explicit-function-return-type': 'error',
             '@typescript-eslint/no-explicit-any': 'warn',
+            '@typescript-eslint/no-unnecessary-type-assertion': 'error',
             '@typescript-eslint/no-unused-expressions': 'warn',
             '@typescript-eslint/no-unused-vars': 'error',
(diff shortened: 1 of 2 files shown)


ferferga:
Then why it's necessary?

3flex:
Sorry I don't understand the question?

The type assertions that were removed are not necessary. An ESLint rule is enabled that detects this (and prevents future violations).

3flex:
Are you asking why the rule needs to be enabled at all?

It's because the recommended type checked rules aren't currently enabled in this project, so I've enabled the rule on its own and fixed the violations.

ferferga:
@3flex And why can't we enable the preset? It's a matter of adding `tseslint.configs.strictTypeChecked` and `tseslint.configs.stylisticTypeChecked`

3flex:
There are currently hundreds of violations if those rule sets are enabled.

Do you want to enable the rulesets then disable all the rules that have violations? Or add eslint ignore rules to each line that triggers a rule on that ruleset?

nielsvanvelzen:
I think enabling a few rules at a time until we have all/most of them and then switch to the preset would be preferred so the pull requests are small.

ferferga (approved):

PR #663, opened 2024-11-02
Don't limit max width for h264 streams to the window width

3flex:
Cast devices variously support video max widths of 1280, 1920 and 3840. `window.innerWidth` returns the max width of the receiver page which is not the same resolution and limits h264 streams to a max of 720p on most devices.

### DIFF
--- a/src/components/codecSupportHelper.ts
+++ b/src/components/codecSupportHelper.ts
@@ -93,23 +93,9 @@ export function getMaxBitrateSupport(): number {
 /**
  * Get the max supported video width the active Cast device supports.
  * @param deviceId - Cast device id.
- * @param codec - Video codec.
  * @returns Max supported width.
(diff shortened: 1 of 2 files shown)


3flex:
@hawken93 the commit introducing this limitation was authored by you on 1/1/21, obviously some time has passed since then so the original issues may have been resolved in CAF or in the Shaka player which this receiver is now configured to use for HLS streams.

I've done limited local testing but this seems to work fine on my Chromecast 2nd gen.

I found various comments saying HLS streaming was problematic when there was a fair bit of dev done a few years back, but I couldn't find specifics, so I'm not sure if whatever problems were seen back then are still an issue today. I'd appreciate any insights here, thank you!

nielsvanvelzen (approved):

PR #680, opened 2024-11-07
Migrate transcoder pinging to jellyfin-sdk-typescript

3flex:


nielsvanvelzen (approved):
In the long term it would be better to keep calling the progress report endpoint while paused instead.

PR #681, opened 2024-11-07
Migrate stopping active encodings to jellyfin-sdk-typescript

3flex:
This function is currently unused so there were no call sites to fix.

Removing this unused function is an option but there's an open TODO item here:

https://github.com/jellyfin/jellyfin-chromecast/blob/2990a569df4bcf2a2dd0587376211babff65343b/src/components/maincontroller.ts#L507-L517

So I thought it was best to leave it in case that's ever addressed.

### DIFF
--- a/src/components/jellyfinActions.ts
+++ b/src/components/jellyfinActions.ts
@@ -7,7 +7,11 @@ import type {
     PlaybackProgressInfo,
     PlayRequest
 } from '@jellyfin/sdk/lib/generated-client';
-import { getMediaInfoApi, getPlaystateApi } from '@jellyfin/sdk/lib/utils/api';
+import {
+    getHlsSegmentApi,
+    getMediaInfoApi,
+    getPlaystateApi
+} from '@jellyfin/sdk/lib/utils/api';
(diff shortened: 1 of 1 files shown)


src/components/jellyfinActions.ts
```diff
@@ -400,21 +400,14 @@ export async function detectBitrate(numBytes = 500000): Promise<number> {
...
- * @param state - playback state.
+ * @param playSessionId - the play session ID to stop encoding
  * @returns Promise for the http request to go through
  */
-export function stopActiveEncodings(state: PlaybackState): Promise<void> {
-    const options = {
+export async function stopActiveEncodings(
+    playSessionId: string
+): Promise<void> {
+    await getHlsSegmentApi(JellyfinApi.jellyfinApi).stopEncodingProcess({
```
nielsvanvelzen, line 409: I think you're missing an import

3flex: I just didn't have more time to fix it during my lunch break :)

nielsvanvelzen (approved):

PR #690, opened 2024-11-10
Remove ItemQuery interface

3flex:


src/helpers.ts
```diff
@@ -550,32 +563,28 @@ export async function getInstantMixItems(
...
-        return {
-            Items: [item],
-            TotalRecordCount: 1
-        };
-    } else {
-        return JellyfinApi.authAjaxUser('Items', {
-            dataType: 'json',
-            query: query,
-            type: 'GET'
+    if (query.ids?.length == 1) {
```
nielsvanvelzen, line 575: This entire branch can be omitted

src/helpers.ts
```diff
@@ -473,26 +475,37 @@ export function getShuffleItems(
...
+    const query: ItemsApiGetItemsRequest = {
+        fields: requiredItemFields,
+        filters: ['IsNotFolder'],
+        limit: 50,
+        recursive: true,
+        sortBy: ['Random'],
+        userId
     };
 
+    let additionalParams: ItemsApiGetItemsRequest;
```
nielsvanvelzen, line 487: I don't see a need for additionalParams, you can just modify the query

nielsvanvelzen (changes requested):

src/helpers.ts
```diff
@@ -473,26 +475,37 @@ export function getShuffleItems(
...
-        Recursive: true,
-        SortBy: 'Random',
-        UserId: userId
+    const query: ItemsApiGetItemsRequest = {
+        fields: requiredItemFields,
+        filters: ['IsNotFolder'],
+        limit: 50,
+        recursive: true,
+        sortBy: ['Random'],
+        userId
```
nielsvanvelzen, line 484: User id doesn't have to be provided as it is inferred from the access token since Jellyfin 10.9

src/helpers.ts
3flex: All properties have the `readonly` keyword applied in `ItemsApiGetItemsRequest` so they cannot be modified after the query is created. But maybe there's a pattern to use here that I'm not aware of...

src/helpers.ts
```diff
@@ -550,32 +562,28 @@ export async function getInstantMixItems(
...
-            {
-                dataType: 'json',
-                type: 'GET'
-            }
-        );
+    const params: ItemsApiGetItemsRequest = {
+        excludeLocationTypes: ['Virtual'],
+        fields: requiredItemFields,
+        limit: query.limit ?? 100,
+        userId
```
3flex, line 571: @nielsvanvelzen I'm guessing `userId` can be dropped here too?

nielsvanvelzen: Yes, basically everywhere where is not required by the SDK it is optional. The only endpoints that still need it are admin related or for the profile picture URL.

src/helpers.ts
nielsvanvelzen: I wasn't aware that the properties are all readonly. I would then write it like so:

```
let query = {...};
// ...
query = {...query, additionalProp: true};
```

So replace the query variable with the additional properties added

nielsvanvelzen (approved):

PR #692, opened 2024-11-10
Migrate calls to authAjaxUser to jellyfin-sdk-typescript

3flex:


src/components/documentManager.ts
```diff
@@ -353,21 +353,19 @@ export abstract class DocumentManager {
...
-            type: 'GET'
+        const response = await getItemsApi(JellyfinApi.jellyfinApi).getItems({
+            imageTypes: ['Backdrop'],
+            includeItemTypes: ['Movie', 'Series'],
+            limit: 1,
+            maxOfficialRating: 'PG-13',
+            recursive: true,
+            sortBy: ['Random']
+            // Although we're limiting to what the user has access to,
+            // not everyone will want to see adult backdrops rotating on their TV.
```
nielsvanvelzen, line 364: Feels like this comment should be above maxOfficialRating

nielsvanvelzen (changes requested):

src/components/documentManager.ts
3flex: Agreed - it was at the bottom in the existing code so I just left it where it was but I'll shift it.

nielsvanvelzen (approved):

PR #694, opened 2024-11-10
Migrate getInstantMixItems to jellyfin-sdk-typescript

3flex:


src/helpers.ts
```diff
@@ -508,38 +518,37 @@ export async function getInstantMixItems(
...
-        UserId: userId
+    if (item.Id === undefined) {
+        throw new Error('Item ID not provided');
+    }
+
+    const query: InstantMixApiRequest = {
+        fields: ['MediaSources', 'Chapters'],
+        itemId: item.Id,
+        limit: 50,
+        userId
```
nielsvanvelzen, line 529: User id doesn't have to be provided as it is inferred from the access token since Jellyfin 10.9

nielsvanvelzen (approved):

PR #713, opened 2024-11-26
Remove abstract keyword for classes that are not subclassed

3flex:


nielsvanvelzen (commented):
I believe these are marked abstract so they cannot be instantiated as they only contain static members. Proper way to fix this would be to not use a class at all (just export members instead). For example see the documentation about the (disabled) ts-eslint rule https://typescript-eslint.io/rules/no-extraneous-class/

3flex:
Yeah I was looking at that as a next step but it needed lots more refactoring.

I'll close this then.

PR #715, opened 2024-11-26
Create interface to hold custom data received from cast sender

3flex:


src/types/global.d.ts
```diff
@@ -106,17 +106,21 @@ declare global {
...
-        itemId: string | undefined;
-        liveStreamId: string | null;
-        mediaSourceId: string | null;
-        playMethod: 'DirectStream' | 'Transcode';
-        playSessionId: string;
-        runtimeTicks: number | null;
-        startPositionTicks: number;
-        subtitleStreamIndex: number | null;
-    }
+    // eslint-disable-next-line @typescript-eslint/no-empty-object-type
```
3flex, line 109: I can't see any way to avoid this.

As an alternative the rule can be configured to ignore empty interfaces that extend from a single interface, but that might not be appropriate.

nielsvanvelzen (approved):

Issue #743, opened 2025-01-04
Chromecast Ultra restricted to 1280x720


olanystrom:
**Describe the bug**

When I try to cast video to my Chromecast Ultra it will request transcode for everything more highres than 1280x720

**To Reproduce**

Cast any video to Chromecast, see transcoding happening, both in video-quality and in FFmpeg.Transcode-logs

**Expected behavior**

Original Video

 Stream #0:0: Video: hevc (Main 10), yuv420p10le(tv, bt2020nc/bt2020/smpte2084), 3840x2160 [SAR 1:1 DAR 16:9], 23.98 fps, 23.98 tbr, 1k tbn (default)
      Metadata:
        BPS             : 70395117
        DURATION        : 02:06:54.691000000
        NUMBER_OF_FRAMES: 182570
        NUMBER_OF_BYTES : 67004633130
        _STATISTICS_WRITING_APP: mkvmerge v89.0 ('And the Melody Still Lingers On (Night in Tunisia)') 64-bit
        _STATISTICS_WRITING_DATE_UTC: 2025-01-03 00:28:47
        _STATISTICS_TAGS: BPS DURATION NUMBER_OF_FRAMES NUMBER_OF_BYTES
      Side data:
        DOVI configuration record: version: 1.0, profile: 8, level: 6, rpu flag: 1, el flag: 0, bl flag: 1, compatibility id: 1

---
Streamed video

Output #0, hls, to '/config/data/transcodes/773641bbac593f7e9ac8b25173250939.m3u8':
  Metadata:
    encoder         : Lavf61.1.100
  Stream #0:0: Video: h264, yuv420p(tv, bt709, progressive), 1280x720 [SAR 1:1 DAR 16:9], q=2-31, 23.98 fps, 90k tbn (default)
      Metadata:
        encoder         : Lavc61.3.100 libx264
      Side data:
        cpb: bitrate max/min/avg: 82347000/0/0 buffer size: 164694000 vbv_delay: N/A
        DOVI configuration record: version: 1.0, profile: 8, level: 6, rpu flag: 1, el flag: 0, bl flag: 1, compatibility id: 1

----
Request headers
![Image](https://github.com/user-attachments/assets/0f795dcc-ea48-4fc4-9faf-98d3cc9d15e6)

Notice the 
  MaxWidth=1280
and
  TranscodeReasons=ContainerBitrateExceedsLimit

headers

**Screenshots**

See Headers above

-   OS: Docker
[...]

Issue #606, opened 2024-09-07
HEVC transcoding Issue on OG Chromecast


alanpaone:
**Describe the bug**

one of my users has an OG chromecast (the 1080p key-shaped one, its otherwise surprisingly fine), and can't play HEVC movies. I found the workaround to turn off tonemapping (would be interested to learn why that works, also), which allows us to sort of play these files, but playback buffers a lot.  These files used to not play, and this is somehow more frustrating. Looking at the FFMPG logs, the resolution is weird, I'm seeing "Height":1606,"Width":3840", the chromecast tops out at 1080p, and is plugged into a 1080p TV. The FPS is probably because I'm using the quicksync on a celeron, and is survivable, but the resolution.....doesn't seem right. 

**To Reproduce**
-Press the Cast Button within Android App
-select OG chromecast
-attempt to watch an HEVC/H265 movie
-Errors and Buffering

**Expected behavior**
This show plays well on my Chromecast Ultra, which supports HEVC. 

**Logs**
Here's the logs from 3 of our attempts to stream the show. There's nothing in the Jellyfin logs around this time, so I didn't both attaching them. 
[FFmpeg.Transcode-2024-09-06_21-39-04_3ebcbbf74a4296d4cd5a8a93a704b17a_4c165e2a.log](https://github.com/user-attachments/files/16919764/FFmpeg.Transcode-2024-09-06_21-39-04_3ebcbbf74a4296d4cd5a8a93a704b17a_4c165e2a.log)
[FFmpeg.Transcode-2024-09-06_21-39-36_3ebcbbf74a4296d4cd5a8a93a704b17a_3cba2049.log](https://github.com/user-attachments/files/16919765/FFmpeg.Transcode-2024-09-06_21-39-36_3ebcbbf74a4296d4cd5a8a93a704b17a_3cba2049.log)
[FFmpeg.Transcode-2024-09-06_21-43-52_3ebcbbf74a4296d4cd5a8a93a704b17a_9742ccc8.log](https://github.com/user-attachments/files/16919766/FFmpeg.Transcode-2024-09-06_21-43-52_3ebcbbf74a4296d4cd5a8a93a704b17a_9742ccc8.log)

solidsnake1298:
[...]

Issue #603, opened 2024-08-28
Cast receiver shouldn't be limited to Home devices


GodTamIt:
**Describe the bug**

The code in [`castDevices.ts`](https://github.com/jellyfin/jellyfin-chromecast/blob/master/src/components/castDevices.ts) and [`codecSupportHelper.ts`](https://github.com/jellyfin/jellyfin-chromecast/blob/master/src/components/codecSupportHelper.ts) make the assumption that the receiving device is one of the Home devices (i.e. `GEN1AND2`, `AUDIO`, `GEN3`, `ULTRA`, `NESTHUBANDMAX`, `CCGTV`). I don't think this is a valid assumption because any Android TV could be a web cast receiver. At least with the latest stable testing, it seems that way on my Android TV. Unless I'm mistaken, this is because the Jellyfin Android TV application hasn't implemented [Cast Connect](https://developers.google.com/cast/docs/android_tv_receiver).

**Additional context**

I used to work on some of these devices 🙃

IDisposable:
Starting to work on this. Cleaning up the device detection first. https://github.com/IDisposable/jellyfin-chromecast/tree/add-generic-android

3flex:
I've been thinking about ways to do this too.

Instead of trying to identify specific device types I think it would be better to dynamically check for required support for resolution, codecs, profiles and levels and build the device profile from there.

For example, instead of trying to identify a specific Cast device then deciding that it supports 720p, 1080p or 4k based on published specs, just check for the resolution via canDisplayType. That way the check should work regardless of device type.

IDisposable:
Couldn't agree more, taking a baby step upright now.

GodTamIt:
> For example, instead of trying to identify a specific Cast device then deciding that it supports 720p, 1080p or 4k based on published specs, just check for the resolution via canDisplayType. That way the check should work regardless of device type.


@3flex I finally got around to implementing something similar to what you suggested in #744. It is a lot more thorough about checking supported video codec profiles, codec levels and resolutions.

Issue #401, opened 2023-04-21
Transcoding bugs


LedyBacer:
I have no way to play 2880x2160 video on chromecast 2015 via official android jellyfin app. Just a black screen instead of video. It seems to me that for some reason the server does not lower the resolution to 1080. 3840x2160 hevc sdr -> transcoded to -> 1080 h264 sdr. But 2880x2160 hevc sdr -> transcoded to -> 2880x2160 h264 sdr. And  even 3840x2160 hevc 60fps transcoded to 1080 h264 60fps stuttering every second, at least because 60fps was not transcoded into 30 fps. Chromecast 2015 supports no more than 30fps at 1080 as i see.

-   OS: Windows
-   Jellyfin version: unstable 10.9.0
-   Cast Receiver version: Stable and Unstable(dosent work at all)
-   Cast client: Official jellyfin android app v2.5.0beta4

LedyBacer:
Oh, so this is all supposed to work on the unstable version? Unfortunately, I can't run any videos on the unstable version.

LedyBacer:
[Log from android app ](https://pastebin.com/EETZd33J)

LedyBacer:
Ok, so i check demo.jellyfin.org use pc with google chrome and that's the only way unstable works

LedyBacer:
I figured out how to make the unstable version work. Instead of connecting by ip, I used to connect to my server by domain via https. Thus, broadcasting via google chrome works on PC and android. Broadcasting the unstable version still doesn't work through the official jellyfin app on android.

PR #744, opened 2025-01-13
Rework device profile building to be more accurate

GodTamIt:
This addresses several issues with how device profile building was done:

  1. Devices were hard-coded, which did not account for the many different device profiles that may not be 1st party Google Chromecast™️ devices. This includes almost every Android TV with Cast functionality built-in.
  2. Different devices may have slightly different profile requirements, so even if an overall codec WAS correct, certain profiles may not be supported.
  3. Resolution constraints were being sent as a poor representation of video codec profile and bit depth constraints. This change fixes this issue.
  4. This also improves the accuracy of the codec detection for less popular codecs like AV1.

Issues:
  - #401
  - #603
  - #606
  - #743


IDisposable:
So much better!

src/components/codecSupportHelper.ts
```diff
@@ -91,119 +295,285 @@ export function getMaxBitrateSupport(): number {
...
+    //
+    // In reality, some encoders may be limited by pixel count instead of
+    // resolution, but other devices may arbitrarily limit the resolution.
+
+    let maxRes = new Resolution(0, 0);
+    const mimeType = videoCodecToMimeType(codec);
+    const codecString = getCodecString(codec, profile, level, bitDepth);
+
+    // Limit the upper bound to 32K, which is more than enough.
+    while (maxRes.width < 30720) {
```
IDisposable, line 323: Technically, 32K is not `30720`, it's 32768... so either the comment above is wrong or the constant here is wrong. Perhaps this 30720 is actually meant to be a some H * W calculation?

src/components/codecSupportHelper.ts
```diff
@@ -91,119 +295,285 @@ export function getMaxBitrateSupport(): number {
...
-        case DeviceIds.CCGTV:
-            return 51;
-    }
-
-    return 0;
+export function getVideoProfileSupport(codec: VideoCodec): string[] {
+    const possibleProfiles = ((): string[] => {
+        switch (codec) {
+            case VideoCodec.H264:
+                return [
```
IDisposable, line 420: These constants and the ones in `getCodecString` are very tightly coupled (as is the logic of formatting the string coupled to the base VideoCodec). Would it be better to actually have a derived class for each VideoCode variant that knows how to format it's own codec string, and exposes the underlying profiles supported?

src/components/codecSupportHelper.ts
```diff
@@ -91,119 +295,285 @@ export function getMaxBitrateSupport(): number {
...
-
-    return '';
+export function getVideoCodecHighestLevelSupport(
+    codec: VideoCodec,
+    profile?: string,
+    bitDepth?: number
+): number | undefined {
+    const possibleLevels = ((): number[] => {
+        switch (codec) {
+            case VideoCodec.H264:
```
IDisposable, line 462: Similar comment to [this](https://github.com/jellyfin/jellyfin-chromecast/pull/744/files#r1917165580), would a specific instance of a derived class for each VideoCodec, that exposed the levels supported as a property be clearer?

src/components/codecSupportHelper.ts
GodTamIt: I'm not sure this is correct. Here are the common 16:9 resolutions:

- **4K:** 3840 × 2160
- **8K:** 7680 × 4320
- **16K:** 15360 × 8640
- **32K:** 30720 × 17280 (speculated, not common)

src/components/deviceprofileBuilder.ts
```diff
@@ -179,89 +177,142 @@ function getCodecProfiles(): CodecProfile[] {
...
+        if (
+            maxLevels.every((l) => l === maxLevels[0]) &&
+            maxBitDepths.every((b) => b === maxBitDepths[0]) &&
+            maxResolutions.every((r) => r.equals(maxResolutions[0]))
+        ) {
+            const maxLevel = maxLevels[0];
+            const maxBitDepth = maxBitDepths[0];
+            const maxResolution = maxResolutions[0];
+
+            const profileConditions = [
```
IDisposable, line 227: You didn't change the logic from before, but this condition seems odd... why do we care if it is anamorphic?

IDisposable (commented):
This looks really good, given how much more computation is going on, are the constraints cached against the play device in any way?

src/components/codecSupportHelper.ts
IDisposable: That's where that number comes from, gotcha... old (16-bit) programmer brain. :)

src/components/codecSupportHelper.ts
GodTamIt: I think I understand what you mean, but do you mind providing a mini example?

FYI, this coupling starts to weaken with an upcoming change I'm stacking on top of this that adds support for checking BT.2020 and SMPTE ST 2084 support.

GodTamIt:
> This looks really good, given how much more computation is going on, are the constraints cached against the play device in any way?

These checks aren't too expensive, as they're mainly just calling into the underlying Chromium engine. Chromium internally should do a lot of the caching here. Because some parts of the output compatibility matrix could, theoretically, be dynamic, I was hesitant to bifurcate the logic into cacheable and non-cacheable profiles. I guess you're right that video tends to be static but I guess resolution isn't always (if you mess around in settings).

Take audio output, for example. It's not uncommon for people to plug and unplug soundbars and/or for them to turn off and on. I do this all the time because of a bug in mine lol.

src/components/codecSupportHelper.ts
GodTamIt: (you are correct that `30720` is a calculation though. it's `1920 * 16`, where `1920` is the width component of 1920 × 1080)

src/components/deviceprofileBuilder.ts
GodTamIt: I don't know enough about what videos are anamorphic anymore (iiuc, it's an artifact of 35mm film -> old CRT 4:3 home displays), so I didn't really change it here. If you have some context as to why we won't care about this anymore, more than happy to remove it.

khsater:
I'm trying to test this but running into a lack of troubleshooting information.  Can you confirm I've configured properly for the test?  I'm running the latest stable version of Jellyfin.

Following contributions.md, I compiled (Debian bookworm) using the pull request for this issue.  I copied the contents of the dist directory to my Apache web server directory.  I configured a Chromecast app and pointed it to the correct Apache directory (note this is reached via NAT-reflection).  I set up a new casting application in system.xml and set my user to use it for casting.

No dice on casting.  The stream does not start.  I see the following repeating in the Chromecast app javascript logs.  I also see some missing images and such in the logs.

```
Valid VP9 codec string requires 4-9 fields, but codecId:vp8 has 1fields.

VM11:16 Valid VP9 codec string requires 4-9 fields, but codecId:vp8 has 1fields.

cast_receiver_framework.js:33 Uncaught (in promise) TypeError: Cannot read property 'slice' of undefined
    at Ob (cast_receiver_framework.js:33)
    at _.ue._.h.canDisplayType (cast_receiver_framework.js:424)
    at O._.h.canDisplayType (cast_receiver_framework.js:721)
    at ar (index-C0He9EXm.js:6)
    at nr (index-C0He9EXm.js:6)
    at fr (index-C0He9EXm.js:6)
    at si (index-C0He9EXm.js:6)
    at Function.<anonymous> (index-C0He9EXm.js:6)
    at Generator.next (<anonymous>)
    at s (index-C0He9EXm.js:1)
```

Do I need to pull in any other pull requests?  Is there any other information that would help track down the cause?

khsater:
Ok, I figured out how to prevent minification and found the following issue with string cases:
image

I'm not familiar enough with JS to know why these are coming back as capitalized strings from VideoCodec, but that seems to be the issue.

src/components/codecSupportHelper.ts
```diff
@@ -45,43 +234,58 @@ export function hasSurroundSupport(): boolean {
...
- * @returns true if HEVC is supported
+ * Checks for every supported video codec.
+ * @returns An array of supported video codecs
  */
-export function hasH265Support(): boolean {
-    return castContext.canDisplayType('video/mp4', 'hev1.1.6.L150.B0');
+export function getSupportedVideoCodecs(): VideoCodec[] {
+    const supportedVideoCodecs: VideoCodec[] = [];
+
+    for (const videoCodec in VideoCodec) {
```
khsater, line 243: The for=in statement returns keys rather than values.  This causes matches in hasVideoCodecSupport to return false.

src/components/codecSupportHelper.ts
```diff
@@ -1,7 +1,196 @@
...
+                    break;
+                default:
+                    profileFlag = 'L';
+                    constraintFlag = 0;
+                    break;
+            }
+
+            // Levels are bound by the luma picture size (total pixels) and
+            // luma sample rate (samples/s).
+            level = level ?? 10;
```
khsater, line 101: Should this be 5?  I didn't trace all the way through to understand where the default value is used, but the levels supported by Google are 5 and 5.1 and the context manager canDisplayType function returned false for my Ultra with this set to 10.

src/components/codecSupportHelper.ts
casebeer: This actually breaks the cast receiver entirely for me on a Chromecast Ultra. 

Passing the enum keys rather than values to `hasVideoCodecSupport()` causes the `videoCodecToMimeType()` and `getCodecString()` lookups to return `undefined`. Passing `undefined` to `castContext.canDisplayType()` causes uncaught exceptions from the Cast SDK. 

I suggest modifying lines 244–245 to be:

        if (hasVideoCodecSupport(VideoCodec[videoCodec])) {
            supportedVideoCodecs.push(VideoCodec[videoCodec]);
        }

nielsvanvelzen:
Hi @GodTamIt are you still interested in finishing this pull request? Thanks

IDisposable:
If not, I would love to carry it forward. We really need this.

src/components/codecSupportHelper.ts
GodTamIt: This was because of an error in the codec string below where the level was being multiplied by 30 instead of 3.

Does the check now pass on your Chromecast Ultra @khsater?

GodTamIt: N.B.: `10` actually represents the human-readable level `1.0` but the codec string doesn't include a period in there.

src/components/codecSupportHelper.ts
```diff
@@ -1,7 +1,196 @@
...
+
+            return `hev1.1.${constraintFlag}.${profileFlag}${level * 3}.B0`;
+        }
+        case VideoCodec.VP8:
+            return 'vp8';
+        case VideoCodec.VP9: {
+            let profileFlag: string;
+
+            switch (profile) {
```
IDisposable, line 110: Might we want a `.toLowerCase()` here for defensive purposes (and change the case literals correspondingly)?

IDisposable (approved):

nielsvanvelzen (approved):

nielsvanvelzen:
Awesome work, thanks!

PR #768, opened 2025-03-04
Skip loading MPL player

3flex:
This app has opted in to use the newer Shaka player for HLS, so there is no need to also load the MPL player which is no longer required.


3flex:
Yes I realised I should have tested locally even though this should be safe to merge. I then moved on to other things. 

I'll close but reopen if I verify later.

PR #888, opened 2025-11-15
Add GitHub Actions workflow for stale issue management

### LINKED ISSUE #879: Configure stale issue action
There are lots of very old issues open. It's hard to tell which are still valid.

I can open a PR to configure the stale issue bot based on the main project's configuration: https://github.com/jellyfin/jellyfin/blob/master/.github/workflows/issue-stale.yml

3flex:
Copied from https://github.com/jellyfin/jellyfin/blob/d1722936c0a453c5b70bca96a8f148b11d2e0688/.github/workflows/issue-stale.yml

Closes #879

### DIFF
--- a/.github/workflows/issue-stale.yml
+++ b/.github/workflows/issue-stale.yml
@@ -1,0 +1,35 @@
+name: Stale Issue Labeler
+
+on:
+  schedule:
+    - cron: '30 1 * * *'
+  workflow_dispatch:
+
+permissions:
+  issues: write
+  pull-requests: write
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
I don't think it makes sense to use this workflow for this repository. We don't have a dedicated maintainer for it and issues are largely ignored, this workflow would just close valid issues.
In addition, the amount of issues created here is very low. So it should be easy enough to go through them manually whenever we have a dedicated person for it.

PR #906, opened 2026-01-08
fix JellyfinApi.jellyfinApi undefined

rienafairefr:
I sometimes experienced weirdnesses on the casting Jellyfin side (like, the android app not connecting or other not identifiable bugs). So I did the only sane thing and built this chromecast receiver from source, containerized, and instrumented it with crash reporting using Sentry.
Not sure I pin pointed yet why the crashes/weirdness I saw happened, but at some point during normal usage I got a javascript crash on using an undefined value, so this patch is guarding against that case which happens, I don't really know why and don't remember (this was a few months ago and I didn't think to send that PR at the time)
Builds with node v22.16.0, npm run lint & npm run test OK

### DIFF
--- a/src/components/maincontroller.ts
+++ b/src/components/maincontroller.ts
@@ -103,7 +103,10 @@ function onMediaElementVolumeChange(event: framework.system.Event): void {
     window.volume = (event as framework.system.SystemVolumeChangedEvent).data;
     console.log(`Received volume update: ${window.volume.level}`);
 
-    if (JellyfinApi.serverAddress !== null) {
+    if (
+        JellyfinApi.serverAddress !== null &&
+        JellyfinApi.jellyfinApi !== undefined
+    ) {
         reportEvent('volumechange', true);
     }
 }


nielsvanvelzen (changes requested):
I don't think this is a proper fix. We should instead look into why the jellyfinApi property is not set when this function is called. Is is perhaps a race condition?

rienafairefr:
From what I can tell in jellyfinApi.ts, 
the jellyfinApi field is set in setServerInfo
```
if (!this.jellyfinApi || regenApi) {
    if (serverAddress && accessToken) {
        this.jellyfinApi = this.jellyfinSdk.createApi(
            serverAddress,
            accessToken
        );
    } else {
        console.error(
            'Server address or access token not provided - could not create instance of Jellyfin API'
        );
    }
}
```
This is called by processMessage, a custom message listener which is aded by 
"window.castReceiverContext.addCustomMessageListener" on line 528 in maincontroller.ts

onMediaElementVolumeChange (where my change is set) is called by an event listener on window.castReceiverContext
which is added in enableTimeUpdateListener(); on line 158 in maincontroller.ts

I can totally see that there is a message (volume update) received after enableTimeUpdateListener is called  but before the "handshake" processMessage is done

althougth the castreceivercontext is only started at the bottom of maincontroller.ts so I don't think my theory works

nielsvanvelzen:
I think a better solution then would probably be to not call enableTimeUpdateListener immediately but instead wait for the server info to be received first.

nielsvanvelzen:
Closing as stale

PR #950, opened 2026-09-20
Stop a codec's video profile entries from contradicting each other

acidg:
## Problem

A codec whose video profiles carry different constraints is announced as one codec profile per video profile, each asserting its own profile as a plain condition:

```
hevc  VideoProfile Equals "main 10"   VideoBitDepth <= 10  VideoLevel <= 186  SDR
hevc  VideoProfile Equals "main"      VideoBitDepth <= 8   VideoLevel <= 186  SDR
hevc  VideoProfile Equals "high 10"   VideoBitDepth <= 10  VideoLevel <= 186  SDR
hevc  VideoProfile Equals "high"      VideoBitDepth <= 8   VideoLevel <= 186  SDR
```

The server applies every codec profile it finds for a codec and requires all of their conditions to hold at once. A stream carries a single video profile, so three of those four can never be satisfied and the codec is never played untouched.

On a Chromecast Ultra, a 1080p HEVC Main 10 file (level 120, 10-bit, SDR) satisfies the first entry and is still re-encoded to H.264:

```
DirectPlay=false. VideoProfile Condition: Equals. ConditionValue: "main"
DirectPlay=false. VideoBitDepth Condition: LessThanEqual. ConditionValue: "8"
DirectPlay=false. VideoProfile Condition: Equals. ConditionValue: "high 10"
DirectPlay=false. VideoProfile Condition: Equals. ConditionValue: "high"
DirectPlay=false. VideoBitDepth Condition: LessThanEqual. ConditionValue: "8"

TranscodeReason=VideoProfileNotSupported, VideoBitDepthNotSupported, ...
```

Every failure comes from an entry describing a different profile. Nothing from the `main 10` entry fails. Devices whose profiles share their constraints take the merged branch above and are unaffected, which is why this is not visible everywhere.

## Fix

Announce the supported profiles once for the codec, as `VideoProfile EqualsAny "main 10|main|high 10|high"`, and gate each per-profile entry on `ApplyConditions` so its constraints apply only to the stream it describes. The shared entry is what still rejects a profile that was never announced, which would otherwise pass unconstrained once no entry applies to it.

## Tests

Both evaluate the announcements the way the server consumes them: an entry takes effect only when its `ApplyConditions` hold, and every entry that takes effect must then be satisfied in full.

- `accepts a stream matching any profile it announced` fails on `main 10` before the change.
- `rejects a stream whose profile it never announced` passes before and after, so the fix cannot be to drop the conditions.

## Verification

On Jellyfin 10.11.11, replaying the generated profile through `/Items/{id}/PlaybackInfo` for that file turns `DirectPlay` and `DirectStream` from false to true. With the device's stereo limit left in place, the server now emits `-codec:v:0 copy` and re-encodes only the 5.1 audio track, where it previously re-encoded the video to H.264.


acidg:
For anyone stuck on this before it ships: the profile can be repaired in front of the server instead, with no change to the receiver.

`POST /Items/{id}/PlaybackInfo` carries the device profile in its body, so routing that one endpoint through a small reverse proxy that applies the same restatement as this PR is enough. With nginx, a regex location outranks the prefix one, which keeps the detour to that endpoint alone:

```nginx
# The repair proxy, which rewrites the body and forwards to Jellyfin.
location ~* ^/Items/[^/]+/PlaybackInfo$ {
    proxy_pass http://127.0.0.1:8099;
}

location / {
    proxy_pass http://127.0.0.1:8096;
}
```

On a Chromecast Ultra the result is identical to running a patched receiver: `DirectPlay` goes from false to true, and where an unrelated audio constraint still forces a transcode, the server emits `-codec:v:0 copy` instead of re-encoding the video.

Two things worth knowing if anyone tries it. Skip entries that already carry `ApplyConditions`, so the repair becomes a no-op once this lands. And forward every HTTP method, not just POST: the receiver is served from another origin, so a preflight `OPTIONS` answered with 501 stops playback outright.

Shadowghost (approved):

Issue #920, opened 2026-05-15
Casting Live TV Resultes in Immediate FFmpeg Failure


clarkent86:
**Describe the bug**

Casting Live TV on unstable Google Cast Version (the only way other media works due to other fixes on the unstable channel) fails. Live TV works fine on Android, Chromium Browser

**To Reproduce**

1. Select Chromecast Device from casting
2. Attempt to play Live TV channel
3. Jellyfin will display the channel loading
4. Jellyfin subsequently reverts back to `Ready to cast`

**Expected behavior**

I expect the Live TV channel to play like other media, and like it does on my Chromium browser/Google TV devices

**Logs**

Stable |

2026-05-15 13:53:20.898758+00:00[08:53:20] [INF] [36] Jellyfin.Api.Helpers.MediaInfoHelper: User policy for <jellyfin-user>. EnablePlaybackRemuxing: True EnableVideoPlaybackTranscoding: True EnableAudioPlaybackTranscoding: True
2026-05-15 13:53:20.970938+00:00[08:53:20] [INF] [29] Jellyfin.LiveTv.LiveTvMediaSourceProvider: Opening channel stream from Emby, external channel Id: hdhr_11.1
2026-05-15 13:53:20.971076+00:00[08:53:20] [INF] [29] Jellyfin.LiveTv.DefaultLiveTvService: Streaming Channel hdhr_11.1
2026-05-15 13:53:20.971136+00:00[08:53:20] [INF] [29] Jellyfin.LiveTv.DefaultLiveTvService: Live stream native_38e2ec422972cf18ae399822dbeeeca8_19e377e4f0b208f3c8bdf8f6c6f6f0fb consumer count is now 6
2026-05-15 13:53:20.971437+00:00[08:53:20] [INF] [36] Emby.Server.Implementations.Library.MediaSourceManager: Live tv media info probe took 0.0004055 seconds
[...]

PR #953, opened 2026-09-21
Stop announcing HLS details the player cannot use

Shadowghost:
Every cast playback on receiver 1.3.1 ends at 0 ms: the receiver loads the stream, never fetches a segment, and the server kills ffmpeg seconds later. It fails on every HLS delivery path with the same server-side ffmpeg command, so the regression is in the load request, not the device profile.

**Changes**
- Drop `hlsSegmentFormat`/`hlsVideoSegmentFormat` from the load request since they describe segments to MPL, and HLS is handed to Shaka via `useShakaForHls`.
- Stop announcing `AC-3` and `E-AC-3` for fMP4 HLS, since `canDisplayType` answers for the platform demuxer rather than the MSE pipeline that plays HLS (would have been too easy...)
- Keep Dolby passthrough on the MP4 direct-play profile, where the platform does demux it.
- Record why the MPEG-TS HLS profile has to exist at all, since live TV is narrowed to it server-side (#920).

@nielsvanvelzen not really sure what to do tbh because of the different render pipelines. Can we not use Shaka?


nielsvanvelzen:
> not really sure what to do tbh because of the different render pipelines. Can we not use Shaka?

🤷 I thought it already used shaka

Shadowghost:
I mean NOT using shaka, not the other way around lol

nielsvanvelzen:
Ooh, so Google wants cast receivers to use Shaka and defaults to it. There's a way to use MPL instead, but that is no longer maintained by them. https://developers.google.com/cast/docs/web_receiver/shaka_migration

